Issue #530, opened 2026-04-17
Embedded docs: mapping, project layout, database, architecture, embedding


PaulSnow:
## Why

After the v1.5.0 packaging work, consumers have a complete tooling story (`dtrules build`, `dtrules verify`, `dtrules docs el/operators/decision-tables/edd/xml-format`) — but five capability areas are not documented in the binary. A downstream repo that imports DTRules cannot answer from `dtrules docs` alone:

- How to structure a project's folders and file naming under the new `_dt`/`_edd`/`_map` convention.
- What a mapping file looks like in XML and xlsx.
- How to design a key/value database whose shape is driven by the EDD.
- How the two data-flow architectures (AI-visible dev vs. deployment) actually plug together.
- How to embed decision-table execution into a Go binary deployment.

## Sub-issues

- [ ] #531 — `dtrules docs project-layout`
- [ ] #532 — `dtrules docs mapping`
- [ ] #533 — `dtrules docs database`
- [ ] #534 — `dtrules docs architecture`
- [ ] #535 — `dtrules docs embedding`

## Acceptance

- Each topic appears in `dtrules docs` index with a one-line description.
- Each topic body is self-contained and has at least one worked example.
- A test asserts every topic renders without error and contains its key section headers.

## Scope note

Documentation-only. No new runtime behavior. If a gap is discovered that requires code, file a follow-up issue rather than bundling.

PaulSnow:
Delivered via PR #542 (docs consolidation) and #551/#552/#553 (EL reference + compiler internals + embedded examples). Closing as shipped; issue is stale housekeeping from the umbrella.

PR #540, opened 2026-04-17
docs: add embedding topic (#535)

### LINKED ISSUE #535: Add 'dtrules docs embedding' topic
Part of #530.

Write the `embedding` doc topic. The goal: a consumer ships **one statically-linked Go binary** that contains every decision table, EDD entry, and mapping. No filesystem dependencies at runtime.

### The embed line

```go
//go:embed rules/xml
var rulesFS embed.FS
```

Document exactly which files get embedded: the compiled `xml/` tree output by `dtrules build`. Not xlsx. Not testfiles.

### Loader from embed.FS

- How to construct a RuleSet / Session from `embed.FS` (not from a filesystem path). Confirm the current API by reading `pkg/dtrules/loader`, `pkg/dtrules/ruleset`, `pkg/dtrules/session` on main before writing. If the loader only accepts a filesystem path today, file a follow-up issue rather than bundling it here — but document the currently working approach (e.g., copy `embed.FS` to a temp dir at startup) and flag it.

### Minimal end-to-end program (~40 lines)

Embeds `rules/xml`, constructs a Session, builds an input entity, calls the entry-point table, reads the result. Self-contained — a consumer could paste it and run.

### Build pipeline

- Dev: `dtrules build` — xlsx ↔ xml round-trip on disk.
- CI: `dtrules verify` — gate before build.
- Binary build: `go build -ldflags="-s -w"` — embedded xml, no xlsx.
- Result: a single binary. No `xlsx`, `xml`, or `testfiles/` on the deployment host.

### Explicit anti-patterns

- Do NOT ship xlsx with the app.
- Do NOT ship the `xml/` tree separately — it goes inside the binary.
- Do NOT read xlsx at runtime.

### Versioning

- `dtrules version` reports the binary version; the embedded rules are pinned to that binary's build.
- If rules need to update without a binary update, that's a separate product decision — document as out of scope here.

## Test

Assert the topic renders and contains `//go:embed` and `embed.FS`.

PaulSnow:
## Summary

- Adds `dtrules docs embedding` topic: how to ship all DTRules rule artifacts in a single Go binary via `//go:embed`
- Registers `"embedding": docEmbedding` in `docs.go` with matching description
- Adds `TestDocumentation_EmbeddingTopic` test verifying `//go:embed` and `embed.FS` appear in the doc

## Content

- What goes into the embed (`xml/` tree only — no xlsx, no testfiles, no `.sync-manifest.json`)
- Current limitation: loader accepts a directory path, not `fs.FS`; workaround uses `os.MkdirTemp` + copy at startup
- ~40-line end-to-end example (embed → load → set entity fields → execute → read result)
- Build pipeline (dev, CI gate with `dtrules verify`, final binary with `-ldflags="-s -w" -trimpath`)
- Anti-patterns, versioning, and binary size expectations (~3.8 MB for TaxReturn xml/, 1–3 MB in binary)

## Follow-up

Filed #536: "Loader should accept fs.FS directly (no tempdir round-trip)" — linked from the doc.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./cmd/dtrules/... ./pkg/dtrules/ -run TestDocumentation` passes
- [x] `TestDocumentation_EmbeddingTopic` asserts `//go:embed` and `embed.FS` appear in the doc

Closes #535, part of #530

### DIFF
--- a/cmd/dtrules/doc_embedding.go
+++ b/cmd/dtrules/doc_embedding.go
@@ -1,0 +1,239 @@
+// Copyright 2024 Paul Snow
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+const docEmbedding = `Embedding DTRules in a Go Binary
+=================================
+
+Goal: ship ONE statically-linked binary that contains every decision table,
+EDD entry, and mapping.  No xlsx files.  No xml/ directory at runtime.
+Copy the binary — it just works.
+
+
+What goes into the embed
+------------------------
+After running 'dtrules build', the xml/ directory in your project contains
+the compiled XML artifacts that the runtime reads.  That is the only thing
+you embed.
+
+    //go:embed rules/xml
+    var rulesFS embed.FS
+
+'rules/xml' is a copy of (or symlink to) your project's xml/ directory,
+placed inside your Go module.  The embed directive captures every file in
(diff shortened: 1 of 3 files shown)


PaulSnow:
Superseded by #542 (consolidated docs topics after docs.go merge conflicts).

PR #865, opened 2026-06-19
feat: rules-reference PDF for SinusitisTherapy

PaulSnow:
Adds a downloadable **Rules Reference PDF** for SinusitisTherapy and links it from the `/sinusitis` hero ("Rules Reference (PDF)").

It's an explanatory document, not just a table dump:

1. **Overview** — what the EDD contains, what decision tables contain, and how the interview UI is generated automatically by *executing* the tables (a rule reaching an unsupplied `collect` field triggers its question).
2. **Entity Data Dictionary** — every entity + attribute with type, default, input/output, and for collected fields the interview question (type, options, lab-style reference range). Collected fields are marked ●.
3. **Decision tables** — all 6 tables as classic grids (conditions Y/N/– on top, actions ✓ below; each numbered column a rule) with EL DSL, comments, policy, contexts, initial actions.

`cmd/tabledoc` is a project-agnostic generator (loads tables via `authoring.OpenProject` + the EDD via `EDD()`), generated from the interactive web-demo rules (`cmd/sinusitis-web/rules`) that carry the collect/question metadata. `make check` passes; website builds clean.


PaulSnow:
Reviewed and brought current (2026-08-07).

**Verdict: sound, merging.** `cmd/tabledoc` builds and its tests pass, the branch merges main cleanly, `make check` is green, and the website builds with the PDF shipping to `dist/` and the hero link resolving.

**The PDF is current.** Worth stating because it was not obvious: the rules under this PR changed in #964 (the SinusitisTherapy repair) while it sat open, which normally means a committed binary has gone stale. It has not. Regenerating from the current rules and diffing the extracted text against the committed PDF comes back **identical** — #964 changed entity `number=` attributes and `<file_metadata>`, none of which tabledoc renders. So the shipped document does describe the repaired rules.

**Added one commit** (19786467) for what that near-miss exposed: nothing would have told us either way. `TestPublishedPDFIsCurrent` digests the rendered HTML — deterministic, so no Chrome needed in CI — and fails with the regeneration command when a rule change reaches the page. Verified it fires by renaming a table. It also records the `--no-pdf-header-footer` flag the shipped PDF was made with; omitting it stamps every page with the date and a `file://` URL, which is how I first noticed my own regeneration was not comparable.

Not covered, deliberately: a change visible only in the PDF (styling, print flags) does not trip the guard.

Issue #864, opened 2026-06-19
EL: add `group … by … summing` (group-by aggregation) + `sum of … where`


PaulSnow:
## Summary
EL expresses every aggregation pattern we need **except group-by-with-sum-on-collision**, which has no
inline construct and no precedent. It's currently only doable as a verbose hand-rolled decision table
(`for all source { for first of result where key==k then add amount, else create+add to }`). Propose a
declarative `group … by … summing …` statement that **lowers to the existing operators** (no VM change),
plus a small `sum of … where` parity fill.

## Motivation
- **Legibility = correctness for rules.** `group payouts by url into recipients summing amount` reads like
  the policy; the hand-rolled `forfirst`/`createentity`/`add to` table (plus key-hoisting to avoid
  same-type entity-stack shadowing) reads like implementation and is easy to get subtly wrong.
- **Recurring pattern.** In our use (Accumulate staking) it appears at ≥3 sites: recipient dedup-by-URL
  (summing amounts), per-type/per-class reward aggregation, and ANAF activity rollups.
- **Asymmetry.** `number of <arr> where <bexpr>` exists, but there is no `sum of <arr> where <bexpr>` — a
  predicated count is sugar yet a predicated sum is not.

## What already works (confirmed in v1.15.0 — no change needed)
- Correlated lookup / join: `for first of arr where outerType.field == …` (top-down entity-stack
  resolution in `interpreter/state.go`).
- Nested iteration; `x is one of arr`; `number of arr where … >= N`; `sum of field in arr`;
  `create T as alias` (landed #812); `add N to field` / `increment`; `sort … by field`; `map … through`.
- So join, nested-count-against-threshold, and entity-building are all expressible today.

## Proposal 1 — `group … by … summing …` (group-by with sum-on-collision)
```
group <sourceArray> by <keyExpr> into <resultColl>
    summing <amountField> [, <amountField> ...]
    [as <ResultEntityType>]
```
**Lowering (sugar over proven ops, no VM op required):** for each source element —
`for first of resultColl where resultColl.<key> == <keyExpr> then { add <src>.<amountField> to <amountField> }
else if none { create <ResultEntityType> as r; set r.<key> = <keyExpr>; set r.<amountField> = <src>.<amountField>; add r to resultColl }`.
Optional future VM operator `groupsum` (hash-keyed, O(n)) if large collections ever matter; the
`forfirst` lowering is O(n²) but fine at our scale.

[...]

Issue #812, opened 2026-05-29
EL: support entity creation in action bodies (or add  form)


PaulSnow:
The `local entity X = new <type> entity` grammar (productions `localEntityUndef`/`localEntityInit`) is only reachable from the `context` non-terminal in the grammar. That means rule authors can declare temp entity locals at table-context-level, but not in `initial_actions` / `actions` bodies — at action-DSL level, the parser errors with "mismatched input 'local' expecting {'action', 'condition', 'policystatement', 'context'}".

This blocks the natural DSL pattern for building arrays of structured entries inside a per-row iteration:

```
contexts:
  for all calculation_context.eligible_accounts

actions:
  -- want this, but `local` is rejected at action level:
  local entity r = new token_recipient entity
  set r.url = payout_url
  set r.amount = net_reward
  add r to calculation_context.recipients
```

The setter / `add` lines DO compile (the postfix is straightforward — `cvs /url xdef`, `r ctx.things swap addto`). Only the `local entity` declaration is rejected at this level.

### Why it matters

Surfaced by the Accumulate staking project's transfer of business logic into DT (umbrella staking/#276). The recipient list (one or two entries per eligible account, depending on delegation) needs a fresh entity per iteration. Without DSL support, the alternatives are:

1. Hand-rolled `action_postfix` override (using `/token_recipient createentity dup entitypush <field> cvs /url xdef ... entitypop calculation_context.recipients swap addto`) — works but bypasses author-friendly DSL.
2. A pre-allocated array of placeholder entities (clunky; needs to know count ahead).
3. Just don't use entity arrays for output — emit parallel scalar arrays. Loses structural typing.

### Suggested DSL surfaces (any of)

1. **Lift `local entity` into action bodies**: a single grammar change that lets `local entity x = <expr>` parse inside `initial_actions` / `actions` — same emitter, scope is the table.
2. **Add an `add new <type> entity to <array>` form** that creates + appends in one step. Subsequent `set <expr>.field = ...` would target the last-appended via a stack convention.
3. **Add an "entity literal" form**: `add token_recipient { url: <expr>, amount: <expr> } to <array>` — JSON-ish constructor.

Option 1 is the smallest grammar change and unblocks the most patterns.

### Workaround in the meantime

`action_postfix` override (v1.14.7's `Action.Postfix` field) lets authors write the postfix directly. Functional but ugly for non-trivial sequences.

Issue #776, opened 2026-05-13
feat(analysis): proper static EDD usage analysis (EL/entity-stack aware, with enumeration-bounded dynamic strings)


PaulSnow:
## Problem

`pkg/dtrules/analysis/edd_unused.go` is a regex pass over DSL text:

```go
var identifierPattern = regexp.MustCompile(`\b([a-z_][a-z0-9_]*)\.([a-z_][a-z0-9_]*)\b`)
```

It only matches dotted `entity.attribute` references. Bare field names that resolve against the runtime entity stack (`for all accounts where identity_url == current_delegate_url` → `identity_url` resolves to `account.identity_url`) are invisible to the analyzer. Consequence: every field accessed via the bare-name path is flagged as "unused EDD field," producing systemic false positives.

A downstream consumer (staking) reports **97 EDD-usage warnings** on their project, the vast majority of which are this class of false positive.

## Scope

Replace the regex pass with proper static analysis over the EL AST. Walk every `*_dt.xml`, parse each DSL fragment with `pkg/dtrules/compiler/el`, and track the entity stack as the walk descends. Resolve bare identifiers against the topmost entity at each reference site.

**Entity-stack pushes the analyzer needs to track:**

- `for all <field>` context — pushes the iterated entity type for the duration of the table.
- `for all <field> where <pred>` / `for first of <field> where <pred>` — same, plus the `where` clause is in scope of the iterated entity.
- `using <entity> { ... }` blocks — pushes a named entity.
- `create <type> as <local>` (#764-era Gap A) — binds a local; field accesses against the local are dotted, but `set local.field = expr` is in scope of the local on the entity stack while the create-block runs.
- `createentity` / `entitypush` postfix idioms in the legacy hand-coded paths — should produce a conservative "unknown entity pushed" marker (see below).
- `perform <Table>` — descends into the called table's context; the analyzer needs cross-table reference tracking.

**Out of scope for the first cut:**

- Run-time deduction of which entity is on the stack inside an arbitrary postfix block. The strict hand-coded-postfix gate (shipped in v1.11.0) means new code can't introduce hand-edited postfix anyway.
- Closure-over-locals analysis (allocate / local@ / local!). Locals are typed at point of definition by the EL compiler; the analyzer can read those type annotations rather than reason about them.

## Dynamic-string corner case (enumeration bounds)

[...]

PR #866, opened 2026-06-19
feat: context-push advisory + sum-of-where (#864)

PaulSnow:
Two changes for decision-table authoring ergonomics.

## 1. Context-push advisory + docs (constants-in-context)
Push a heavily-referenced entity onto the stack once at the entry table; the stack propagates down `perform` calls, so every reachable table reads its fields unqualified.
- `analysis.SuggestContextPushes`: flags an entity referenced with a qualifier ≥3× in tables where it's not on the effective stack (own contexts ∪ inherited via the #776 call-graph propagation); detects existing `add X to context` pushes (goes silent once adopted); targets the suggestion at the call-graph root. Surfaced as `context_hints` in `dtrules review` — advisory only.
- Embedded docs: "Shared Constants: Push Once, Reference Unqualified" in `dtrules docs decision-tables`.

## 2. `sum of … where` predicated sum (#864, Accumulate staking)
Parity fill for the existing `number of … where`.
- New EL.g4 alternatives `intSumOfWhere` / `floatSumOfWhere`; parser regenerated (ANTLR 4.13.1). Lowers to a `forall` fold gating each addition on the predicate (`0 arr { bexpr { iexpr + } if } forall`) — no VM op.
- Docs in `dtrules docs el` + `el-reference.md`; marks `el_grammar_gaps.md` **Gap A RESOLVED** (`create <type> as <local>`, #812).

`make check` passes. New tests: advisory (root/silent/threshold) and `sum of … where` (gated fold vs unconditional). The #864 group-by (P1) and same-type-join alias are deferred to a follow-up per scope.


PaulSnow:
Depends on #868 (issue #867): this branch carries the `sum of … where` operand-order fix for its own new emitters, but reverts to the pre-#867 plain-fold emitters. Rebase onto main after #868 merges to pick up the `sum of` / `number of … where` fix.

PR #871, opened 2026-06-20
fix(ci): stop Go Tests runner kill (exit 143) by short-circuiting the 50s load test (#870)

### LINKED ISSUE #870: CI: Go Tests job killed with exit 143 (SIGTERM) on every run, incl. main
## Summary
The **Go Tests** workflow ("Build and Test (1.24)") fails on essentially every run — including pushes to `main` — with **exit code 143 (SIGTERM)** and `##[error]The runner has received a shutdown signal`. The job is killed ~80–130s in, during the test step, not at a `--- FAIL`. CI is effectively always-red for Go, so it provides no signal and forces merging over red checks.

## Where
- Workflow: `.github/workflows/go-tests.yml`, job `Build and Test`, step **"Run scoped build + tests (matches `make check`)"**.
- `make check` ends with `go test -count=1 ./...` (the **whole module**, including the heavy root `pkg/dtrules/` integration tests, interpreter, `runtime/nativeasm`, web/interview, sample-project loads). That whole-module run is what gets terminated.
- `runs-on: ubuntu-latest` (GitHub-hosted, ~7 GB RAM, 2 vCPU).

## Evidence (consistent across branches)
Recent `go-tests` runs — failure on `main` and every PR branch:
- `main` — runs 27860834823, 27860863737
- `feat/constants-in-context-advisory` — 27857731971, 27857288772, 27842769336, 27841696440 …
- `fix/issue-867-…` — 27857651084
- `feat/decision-table-pdf` — 27857365182, 27838977263, 27837687822 …

All show `Process completed with exit code 143` + "runner has received a shutdown signal". Local `make check` passes, so this is the CI runner, not the tests themselves.

## Likely cause
Exit 143/SIGTERM with "runner received shutdown signal" on a hosted runner usually means the runner VM was terminated — most often **memory pressure** during `go test ./...` (which builds and runs every package's test binary, several concurrently by default) on a 7 GB runner, or a single heavy/leaky/process-spawning test. The consistent timing (~near the end of the run) argues for something deterministic rather than random preemption.

## Suggested investigation / mitigations
- Reproduce with diagnostics: a temporary CI step running `GODEBUG=gctrace=1 go test -v -count=1 ./...` and printing `free -m` / `dmesg | tail` on failure to confirm OOM vs hang.
- Serialize package test binaries to cut peak memory: `go test -p 1 ./...` (and/or `-parallel 1`).
- Split the suite into per-area jobs (e.g. `pkg/dtrules/` root + `runtime/...` + the rest) so no single job holds everything in memory.
- If memory-bound, bump to a larger runner (`ubuntu-latest-4-core` / `-8-core`).
- Add `timeout-minutes:` to the job so a hang fails fast and visibly instead of as a vague shutdown.

## Impact
Every PR this cycle (#863, #865, #866, #868) showed a red "Build and Test" that was purely this SIGTERM, forcing maintainers to merge over red and eroding trust in the gate. Worth fixing so CI is a real signal again.

PaulSnow:
Fixes #870.

## Diagnosis
The "Build and Test" job was killed with **exit 143 (SIGTERM, "runner has received a shutdown signal")** on every run, including `main` — not a `--- FAIL`.

- **Memory ruled out:** max single test-binary heap is **146 MB** (via `GODEBUG=gctrace=1`), so it's not OOM; `-p 1`/`GOMEMLIMIT` would not have helped.
- **Real cause:** one test, `TestBackwardCompatibilityFallback`, takes **~50s** — 93% of the root `pkg/dtrules/` package and ~⅓ of the whole-module test time — because it loads the full **523-table** TaxReturn project and builds every decision-table interpretation tree (`processCol` + pairwise `EqualsNode`, GC-dominated). On the 2-vCPU hosted runner that stretches the job toward the ~130s window where it gets killed.

## Fix
- Guard that test behind `testing.Short()` — it still runs in a full `make check` (local + release).
- `Makefile`: `go test -count=1 $(GOTESTFLAGS) ./...` so CI can pass `-short` without forking the gate.
- `go-tests.yml`: `make check GOTESTFLAGS=-short` + `timeout-minutes: 15` (a true hang now fails clearly instead of a vague shutdown).

**Effect (measured locally):** fast-gate test phase drops from ~70s to **~8s** (root package 52s → 1.1s). This PR's own CI run is the validation — Build and Test should now go green.

## Follow-up
The ~50s load itself looks like an **O(n²) in the decision-table tree builder** (the `EqualsNode` dedup). Optimizing it would speed the full suite and release too — worth a separate issue.


PaulSnow:
Superseded by #873. That branch archives `multifile_integration_test.go` (the 50s TaxReturn-loading test `TestBackwardCompatibilityFallback`) behind `//go:build archive`, so the test no longer runs in a default `make check` at all — which removes the SIGTERM runner-kill (#870) from **both** `go-tests.yml` and `release.yml`, not just the fast gate. `-short` is therefore unnecessary, and the two PRs conflict on the same test file.

The one hygiene bit worth keeping — a job `timeout-minutes` so a true hang fails clearly instead of silently skipping "Create GitHub Release" — has been carried into #873 (applied to `release.yml`). Closing this as superseded; #870 is resolved by #873.

Issue #879, opened 2026-06-25
EL codegen: no cross-generator parity test; symbol-map divergences


PaulSnow:
## Summary
#874 was caused by two DSL→postfix generators disagreeing on identical input. Nothing guards against the next such divergence, and the audit found more.

## Three generators, divergent symbol maps
- Authoring Save — `authoring/project.go:loadEDD` — non-recursive `Glob`, **case-preserved**.
- Workbook import — `excel/workbook_importer.go:eddSymbols` — **lowercases** entity/field keys (line ~110).
- CLI build — `cmd/dtrules/edd_symbols.go:loadEDDSymbols` — recursive `WalkDir`, case-preserved.

### Latent bugs from the divergence
1. **Case-sensitivity (new #874-class bug):** the workbook importer lowercases EDD keys while the other paths preserve case. A mixed-case field (`Patient.Age`) would miss the symbol lookup on the workbook path only → integer ops instead of typed ops. Unverified but plausible.
2. **EDD discovery scope:** authoring uses non-recursive `Glob`; build uses recursive `WalkDir`. Nested `xml/states/*_edd.xml` is invisible to authoring Save.

## What's missing
- **No cross-generator parity test.** Need a harness that compiles a DSL corpus through all three paths and asserts identical postfix **and** identical runtime results. This would have caught #874 and would catch the case-sensitivity divergence.
- Token-presence tests can't catch operand-order / op-selection bugs (the reason #867 and #874 slipped). Standardize on execution/round-trip tests.

## Fix direction
Unify symbol-map construction (one shared builder: consistent case handling + recursive discovery) and add the parity + execution harness.

Issue #867, opened 2026-06-20
EL: forall-fold aggregations emit operands in wrong order (runtime crash)


PaulSnow:
## Summary
The `forall`-fold aggregation emitters produce operands in the wrong order and **throw at runtime**. Affects the shipped `sum of <x> in <arr>` and `number of <arr> where <bexpr>` (and, on the feature branch, the new `sum of … where`).

## Root cause
Two stack-discipline mismatches, both verified by execution:

1. **forall**: `opForall` is `( body array -- )` — it pops the **array off the top**, then the body. The emitters emit `seed arr { body } forall`, leaving the body block on top, so `forall` takes the block as the array and iterates its tokens → `[typecheck] Forall: Encountered a non-Entity entry in array`.
2. **if** (the `where` variants): `opIf` is `( body boolean -- )` — pops the **boolean off the top**. The emitters emit `bexpr { body } if`, so `if` reads the block as the boolean → `BooleanValue: No Boolean value exists`.

Correct order matches the working `VisitForallWhere` template: `seed { { body } bexpr if } arr forall`.

## Why it was never caught
The only execution test for these is `t.Skip`-archived (#520); the feature tests assert token *presence*, which can't see operand order. Verified with a fresh execution probe: `0 kids { tax + } forall` errors; `0 { tax + } kids forall` returns the right sum.

## Affected emitters (pkg/dtrules/compiler/el/postfix_emitter.go)
- `VisitIntSumOf`, `VisitFloatSumOf` — plain `sum of`
- `VisitIntNumberOfWhere` — `number of … where`
- (feature branch #866/#864) `VisitIntSumOfWhere`, `VisitFloatSumOfWhere`

## Very likely also affected (separate follow-up)
The boolean aggregations share the `seed arr { body } forall` shape and almost certainly have the same bug: `VisitBoolThereIsInArrayWhere`, `VisitBoolThereIsNoInArrayWhere`, `VisitBoolAllHave`, `VisitBoolOneOfHasa`, `VisitBoolMatchForall`. Needs its own execution tests.

## Fix
Reorder the affected emitters to `seed { body } arr forall` (and `{ block } bexpr if` for the predicated ones). Add an **execution** test that compiles EL and runs it over real entities, asserting numeric results.

Found via a deep review of the #865/#866 branches.

PR #880, opened 2026-06-25
fix(el): correct forall operand order in boolean aggregations (#877)

### LINKED ISSUE #877: EL codegen: forall boolean aggregations have reversed operand order
## Summary
Follow-up to #867. The boolean-aggregation constructs emit `seed <array> { body } forall` (array before body), violating `opForall = ( body array -- )` — the identical crash #867 fixed for the numeric folds. The #867 commit explicitly flagged these as "likely the same bug."

At runtime `forall` pops the array off TOP, so it takes the **body block** as the array → `"non-Entity entry in array"` crash.

## Affected constructs (postfix_emitter.go)
| construct | Visit method | line |
|---|---|---|
| `all <arr> have …` | `VisitBoolAllHave` | 4945 |
| `one of <arr> has a …` | `VisitBoolOneOfHasa` | 4956 |
| `there is <arr> where …` | `VisitBoolThereIsInArrayWhere` | 4998 |
| `there is no <arr> where …` | `VisitBoolThereIsNoInArrayWhere` | 5009 |
| `there is MATCH FORALL …` (inner forall) | `VisitBoolMatchForall` | 5310 |

Correct order: `seed { body <op> } <arr> forall` (op = `and`/`or`). The entity-scoped `there is … where` variants (4967–4987) don't use forall and are fine.

## Tests
None of these have execution tests (only the two #867-fixed numeric folds do). Fix the whole family together and add one execution test mirroring `TestForallFoldExecution`.

PaulSnow:
Fixes #877. Follow-up to #867.

## Bug
The boolean-aggregation constructs emitted `seed <array> { body op } forall`, but `opForall` is `( body array -- )` and pops the array off the top — so `forall` iterated the **body block** as the array → runtime crash `"non-Entity entry in array"`. Identical to the bug #867 fixed for the numeric folds; the #867 commit explicitly flagged these as the likely twin.

## Fix
Reordered all five emitters to put the block before the array (`seed { body op } <arr> forall`), matching the working `number of … where` / `sum of … in` template:

- `VisitBoolAllHave`
- `VisitBoolOneOfHasa`
- `VisitBoolThereIsInArrayWhere`
- `VisitBoolThereIsNoInArrayWhere`
- both folds of `VisitBoolMatchForall` (outer + inner)

Reordering tokens does not change runtime entity-stack depth (forall pushes each element while running the body), so `MatchForall`'s `entityfetch 1` still resolves the outer element.

## Test
`TestBooleanAggregationExecution` compiles `all … have` / `one of … has a` and **runs** them over real entities, asserting boolean results. It reproduces the exact `"non-Entity entry in array"` crash on the old code and passes with the fix. Token-only tests can't catch this — the tokens are identical regardless of order.

Coverage note: the two array `there is … where` variants and `MatchForall` require `eexpr`/two-array setups not in this harness; they share the identical one-line reorder with the two execution-tested constructs. A broader cross-construct execution harness is tracked under #879.

`make check` green.

### DIFF
--- a/pkg/dtrules/boolean_aggregation_exec_test.go
+++ b/pkg/dtrules/boolean_aggregation_exec_test.go
@@ -1,0 +1,123 @@
+// Copyright 2026 Paul Snow
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dtrules_test
+
+import (
+	"testing"
+
+	"github.com/DTRules/DTRules/pkg/dtrules"
+	"github.com/DTRules/DTRules/pkg/dtrules/compiler/el"
+	"github.com/DTRules/DTRules/pkg/dtrules/entity"
+	"github.com/DTRules/DTRules/pkg/dtrules/interpreter"
+	"github.com/DTRules/DTRules/pkg/dtrules/session"
+)
+
+// TestBooleanAggregationExecution exercises the forall boolean-aggregation
+// constructs (#877, the follow-up #867 explicitly flagged) end to end. Like the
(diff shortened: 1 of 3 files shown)


PaulSnow:
Coverage update: added `TestMatchForallOperandOrder` (exact-postfix order assertion for the nested `match for all` folds). Investigation found the `there is … in … where` **array** variants (boolThereIsInArrayWhere/NoInArrayWhere) are **unreachable** — the entity alternatives shadow them in the grammar — so their reorder is defensive only. Reachable forall aggregations now covered: `all … have` + `one of … has a` (execution) and `match for all` (exact postfix).

Issue #874, opened 2026-06-24
Authoring SDK Save() emits integer ops (`- cvi`) for fixed-point operands — #790 fix not applied to the Save/syncToXML path


PaulSnow:
## Summary

#790 fixed `dtrules compile` to emit fixed-point operators (`fp-`, `fp*`, `fp/`, `cvfp`) for `fixed`-typed operands. But the **authoring SDK's `Project.Save()`** path still emits **integer** operators (`-`, `cvi`) for the same fixed-point DSL.

Since #817 ("postfix always compiled from DSL during Save"), `Save()` is the canonical postfix generator for SDK-based authoring tools. So re-saving an action that was previously correct **silently degrades** its postfix and breaks the runtime / loses precision.

The two postfix generators now **disagree** on identical input: `dtrules compile` is correct, `Save()` is not.

## Isolated repro

Take any table with a `fixed - fixed` action. EDD declares both operands `type="fixed"`:

```xml
<field name="supply_limit"  type="fixed" .../>
<field name="acme_issued"   type="fixed" .../>
<field name="unissued_supply" type="fixed" .../>
```

Open the project via the authoring SDK and `UpdateAction` the action with its **own existing DSL** (no change), then `Save()`:

```go
tbl := p.Table("Calculate_Budget")
// action 1 DSL is, unchanged: "set unissued_supply = supply_limit - acme_issued"
tbl.UpdateAction(1, authoring.Action{DSL: a.DSL, Comment: a.Comment})
p.Save()
```

Resulting postfix:

| generator | postfix for `set unissued_supply = supply_limit - acme_issued` |
|---|---|
| committed / `dtrules compile` (correct) | `supply_limit acme_issued fp- cvfp /unissued_supply xdef` |
| **authoring SDK `Save()` (bug)** | `supply_limit acme_issued - cvi /unissued_supply xdef` |

## Impact

`-` is integer `opSub` and `cvi` is convert-to-int; for `fixed` (`*RFixed`) operands the runtime either rejects with `[Conversion Error] IntValue: No Integer value exists for this type` (same failure mode as #790) or, when an upstream double is involved, the `cvi` truncates the result (e.g. a nanoACME-precision budget collapses to whole units).

Worse, it cascades: a downstream multiply that consumes `unissued_supply` then compiles to `fmul` (float) and a later `fphalfup/` aborts with `RFixed.promote: cannot promote double to fixed implicitly`.

[...]

Issue #790, opened 2026-05-24
`dtrules compile` emits integer operators (`min`, `-`, `*`, `/`) for fixed-point operands — runtime rejects on `IntValue`


PaulSnow:
## Summary

`dtrules compile` emits **integer operators** (`+`, `-`, `*`, `/`, `min`, `max`) for `fixed`-typed operands where the EL/DSL clearly references fixed-point fields. The compiled postfix then **fails at runtime** under the v1.14.x strict loader because the integer ops can't read a `RFixed` value:

```
[Conversion Error] IntValue: No Integer value exists for this type
PostFix: effective_withholding withholding_cap withholding_paid -  ERROR==> min  <== cvi /effective_withholding xdef
```

The same DSL compiled and executed correctly under v1.12.0's in-loader EL compiler (the path #785 removed). The standalone `compiler/el` reachable from `dtrules compile` produces postfix that the runtime rejects.

## Repro

EDD declares the fields as `fixed`:

```xml
<field name="effective_withholding" type="fixed" .../>
<field name="withholding_cap"       type="fixed" .../>
<field name="withholding_paid"      type="fixed" .../>
```

DT action:

```xml
<action_dsl>set effective_withholding = the minimum of effective_withholding and (withholding_cap - withholding_paid)</action_dsl>
<action_postfix></action_postfix>
```

`dtrules compile` fills:

```xml
<action_postfix>effective_withholding withholding_cap withholding_paid - min cvi /effective_withholding xdef</action_postfix>
```

Runtime crashes because:

- `-` is `opSub` (integer subtraction, `pkg/dtrules/operators/math.go:30`)
- `min` is `opMin` (integer min, `pkg/dtrules/operators/math.go:68`)
- both call `.IntValue()` on their operands; the operands are `*RFixed`, so the call returns "No Integer value exists for this type"

The expected operators for `fixed` operands exist and are properly registered (`pkg/dtrules/operators/fixed.go:26-42`):

```
fp+    opFixedAdd
fp-    opFixedSub
fp*    opFixedMul
fp/    opFixedDiv
fpmin  opFixedMin
fpmax  opFixedMax
```

When I hand-edit the postfix to use them:

```xml
<action_postfix>effective_withholding withholding_cap withholding_paid fp- fpmin /effective_withholding xdef</action_postfix>
```

[...]

PR #881, opened 2026-06-25
fix(el): reject implicit double/exact-type mix (#876)

### LINKED ISSUE #876: EL codegen: double operand un-cast in mixed fixed/bigint arithmetic
## Summary
The EL→postfix emitter drops the cast on a `double` operand when the other operand is `fixed` or `bigint`, so the double reaches a fixed/bigint op un-cast and the runtime rejects it. Same bug class as #790 / #874.

## Evidence (compiled, postfix read directly)
| DSL (`fx`=fixed, `db`=double, `bi`=bigint, `it`=integer) | emitted postfix |
|---|---|
| `set fx = fx2 + db` | `fx2 db fp+ cvfp …` ← `db` un-cast before `fp+` |
| `set bi = bi2 + db` | `bi2 db b+ cvbi …` ← `db` un-cast before `b+` |
| `fx > db` | `fx db fp>` ← `db` un-cast |
| `set fx = fx2 + it` (control) | `fx2 it cvfp fp+ …` ← integer **is** cast |

`promoteArithType` (`postfix_emitter.go:282`) makes fixed/bigint win, but `emitTypeCast` (`postfix_emitter.go:347`) only handles int→bigint, int→fixed, bigint→fixed — never double→fixed or double→bigint. At runtime `fp+` hits the promote guard `fixed.go:188`: *"cannot promote double to fixed implicitly"*.

Affects every arithmetic + relational op for `fixed×double` and `bigint×double`.

## Open design question
`promoteArithType` ranks **Fixed > BigInt > Double > Integer**, so `bigint × double` promotes to **bigint** and would truncate the double's fraction. Casting double→fixed (cvfp) is lossless-intent and clearly right; double→bigint truncation may be the wrong promotion — decide whether bigint×double should instead promote to double/fixed or be rejected.

## Also (related)
Name-path `double == double` compiles to `streq` (string compare) instead of `f==` — inconsistent with the iexpr path's `f==`. Documented as intentional grid-snap avoidance, but untested and surprising.

## Tests
No execution coverage for any mixed double combination. Needs a compile→run→assert-value test.

PaulSnow:
Fixes #876.

## Bug
Mixing a `double` with an exact type (`fixed`/`bigint`) emitted the wider type's op with the **double operand un-cast**:

| DSL | old postfix | runtime |
|---|---|---|
| `set fx = fx2 - db` | `fx2 db fp- …` | ✗ "cannot promote double to fixed implicitly" |
| `set bi = bi2 + db` | `bi2 db b+ …` | ✗ IntValue on RDouble |
| `fx > db` | `fx db fp>` | ✗ |

Same class as #790 / #874: `promoteArithType` makes the exact type win, but `emitTypeCast` never cast `double→fixed`/`double→bigint`.

## Decision: reject, don't auto-cast
Per maintainer call, the compiler now **rejects** the implicit mix instead of silently casting — consistent with the runtime guard (`RFixed.promote`) that already refuses implicit `double→fixed`. Silent auto-cast would reintroduce exactly the precision loss that guard exists to prevent. Authors opt in with an explicit cast.

```
set budget = supply - rate     // supply: fixed, rate: double
  emission errors: cannot combine double with fixed implicitly;
  add an explicit cast (e.g. "(fixed) x" to keep exactness,
  or "(double) x" to compute in double)
```

## Changes
- `e.promote(a,b)` — emission-site wrapper over `promoteArithType` that flags a double/exact mix on every arithmetic, ordering-comparison, and min/max site. Bare `promoteArithType` stays for type inference (no false positives on nested subexpressions).
- Name-path `==`/`!=` — the numeric dispatch skips `double` and would emit a meaningless `streq` for `fixed/bigint == double`; it now gets the same reject.
- `CompileAction` — stops masking an emission error with the raw-statement fallback's parse error. A clean-parse/failed-emission is definitive (sentinel `errEmission`), so the real diagnostic surfaces instead of `mismatched input 'set'`.

## Tests
- `TestMixedDoubleExactRejected` — `+ - * /`, `min`, and comparisons across fixed/bigint × double all reject with the cast guidance.
- `TestMixedDoubleExactAllowedForms` — explicit casts and the legitimate int/double widening cases still compile (guards against over-rejection).

`make check` green — no existing test or sample project relied on the old behavior.

### DIFF
--- a/pkg/dtrules/compiler/el/compiler.go
+++ b/pkg/dtrules/compiler/el/compiler.go
@@ -24,11 +24,19 @@
 package el
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/antlr4-go/antlr/v4"
 )
+
+// errEmission marks a failure that occurred after a clean parse, while the
+// emitter was producing postfix (e.g. a type error). CompileAction uses it to
+// distinguish a definitive emission failure from a parse-structure mismatch:
+// the former must not trigger the raw-statement fallback, which would reparse
+// and report a confusing parse error that masks the real diagnostic.
+var errEmission = errors.New("emission errors")
 
 // Compiler compiles Expression Language (EL) to postfix notation.
 type Compiler struct {
@@ -119,6 +127,13 @@ func (c *Compiler) CompileAction(el string) (string, error) {
 		return result, nil
 	}
 
+	// A clean parse that failed only at emission (e.g. a type error) is
+	// definitive — don't fall back to the raw-statement attempt, which would
+	// reparse and report a confusing parse error that hides the real one.
+	if errors.Is(err, errEmission) {
+		return "", err
+	}
+
 	// Try without action prefix (might be a raw statement)
 	result, err = c.compile(el)
 	if err == nil {
@@ -202,7 +217,7 @@ func (c *Compiler) compile(el string) (string, error) {
 		for i, err := range errs {
 			errStrs[i] = err.Error()
 		}
-		return "", fmt.Errorf("emission errors: %s", strings.Join(errStrs, "; "))
+		return "", fmt.Errorf("%w: %s", errEmission, strings.Join(errStrs, "; "))
 	}
 
 	return c.emitter.Emit(), nil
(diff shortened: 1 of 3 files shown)


PaulSnow:
Scope note: this rejects the implicit double/exact mix on the **binary arithmetic + comparison** paths. The **field-mutation** (`add/subtract … to/from`) and **multiply/divide-by** paths still silently auto-cast double→exact (verified: they snap, not crash) — tracked as #882, which needs a value-type-threading refactor across the mutation visitor family. Test now also covers max, strict `<`, and the iexpr-path `==`/`!=`.

PR #908, opened 2026-07-25
feat(excel): entity numbers + number backfill for entities and tables

PaulSnow:
Entities gain an optional `number` attribute ordering them the way `TABLE_NUMBER` orders decision tables. Both canonical XML emitters normalize numbers on every write, mirroring the provenance backfill from bf6a4db3: numbered elements keep their numbers; unnumbered ones get document-order numbers in increments of 100 continuing above the highest existing (100-grid aligned, starting at 100 from scratch).

Unit tests pin backfill around existing numbers, from-scratch numbering, and continuation above a non-aligned maximum.

First of a 3-PR stack: this → #apiserver → #embedded-editor.

### DIFF
--- a/pkg/dtrules/excel/dt_importer.go
+++ b/pkg/dtrules/excel/dt_importer.go
@@ -513,7 +513,29 @@ func normalizeProvenance(tables *DecisionTablesXML, xmlPath string) {
 	}
 }
 
+// normalizeTableNumbers backfills missing TABLE_NUMBERs. Tables that already
+// carry a numeric number keep it; the rest are assigned numbers in document
+// order, in increments of 100, continuing above the highest existing number
+// (starting at 100 when none are numbered).
+func normalizeTableNumbers(tables *DecisionTablesXML) {
+	max := 0
+	for i := range tables.Tables {
+		if n, err := strconv.Atoi(strings.TrimSpace(tables.Tables[i].AttributeFields.TableNumber)); err == nil && n > max {
+			max = n
+		}
+	}
+	next := (max/100)*100 + 100
+	for i := range tables.Tables {
+		t := &tables.Tables[i]
+		if _, err := strconv.Atoi(strings.TrimSpace(t.AttributeFields.TableNumber)); err != nil {
+			t.AttributeFields.TableNumber = strconv.Itoa(next)
+			next += 100
+		}
+	}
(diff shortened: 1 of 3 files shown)


PaulSnow:
## Review — feat(excel): entity numbers + number backfill

**Overview**: Adds an optional `number` attribute to EDD entities and backfills missing entity/table numbers (increments of 100, document order, continuing above the current max) in both canonical XML emitters. Small, focused, well-tested at the unit level. **Verdict: good to merge** after the notes below; the one defect found was already fixed on the branch.

**Correctness**
- ✅ Backfill math is right: `(max/100)*100 + 100` continues cleanly above non-aligned maxima (pinned by the 1550→1600 test), and a single pass can't collide with itself because assignments are strictly above the pre-scan max.
- ✅ Ordering interplay: table-number backfill runs *before* provenance normalization, so the provenance pass's "non-numeric after numeric" append ordering effectively vanishes — every table is numeric by then. Benign, and arguably simpler, but it does mean sheet-number assignment for previously unnumbered tables is now numeric-ordered rather than name-ordered. No action needed; worth knowing.
- ⚠️ Pre-existing duplicates are preserved: two entities both authored as `300` stay `300`. That's consistent with "keep what's authored," but a `dtrules review` advisory for duplicate entity/table numbers would close the loop. Suggest a follow-up issue.
- 🔧 **Fixed during review**: inserting `normalizeEntityNumbers` orphaned `WriteXML`'s doc comment onto the new function. Restored in a follow-up commit on this branch.

**The real gap (follow-up issue recommended)**: entity numbers exist only in the XML surface. The EDD *Excel* sheet has no Number column, so under the authoring contract (Excel is the system of record) a user-assigned entity number will not survive an Excel-driven `dtrules build` — it'll be regenerated by document order. Deterministic, so nothing breaks silently today, but explicit numbers set through the API are only durable until the next Excel round-trip. The fix belongs in the EDD sheet exporter/importer and deserves its own PR.

**Tests**: unit coverage of both normalizers is good (backfill-around-existing, from-scratch, grid continuation). Missing: a `WriteXML`-level pin that the emitted file carries the numbers (the normalizers are tested directly; the wiring into WriteXML is only exercised transitively). Cheap to add alongside the provenance tests' pattern.

Issue #757, opened 2026-05-02
cmd/dtrules and cmd/api/main.go duplicate engine wiring — extract pkg/dtrules/sdk


tradesmantools:
## Summary
There is no shared SDK package. `cmd/dtrules` (`cli.go:43-75`) and `cmd/api/main.go` each glue the engine pipeline together independently — `excel.WorkbookImporter`, `sync.Syncer`, compiler, interpreter, session. If a third caller appears (or the existing two drift), behavior diverges silently.

## Context
CLAUDE.md already advertises `pkg/dtrules/sdk` with `engine.NewContext` / `engine.Execute` — but the directory does not exist (filed separately as the CLAUDE.md drift issue). The doc claim is the right shape, the package just was never created.

## Suggested fix
Extract the common wiring into `pkg/dtrules/sdk/engine.go`:
- `NewContext(opts ...Option) (*Context, error)` — owns importer + compiler setup
- `Execute(ctx *Context, input map[string]any) (*Result, error)` — runs the pipeline end-to-end
- `cmd/dtrules` and `cmd/api` switch to importing `sdk`

This also unblocks the embedded-use-case story that CLAUDE.md already promises.

---
Reported by codebase audit on `local/work` @ 8d4d3bb1.

PR #909, opened 2026-07-25
feat(apiserver): extract the API server into an importable package

PaulSnow:
Moves the UI's REST backend from `cmd/api` (a 2,400-line main package) to `pkg/dtrules/apiserver` so `dtrules edit` can embed it; `cmd/api` becomes a flag-parsing shim. First installment of the #757 SDK extraction — the load-parity test moves with the code.

Beyond the move, this carries the session's API correctness work:

- **Parse the modern XML surface** the old structs predated: `condition_dsl`/`action_dsl`/`context_dsl`, structured `<initial_action>` blocks (previously silently dropped), and canonical policy statements (`policy_statements>policy_statement` with `policy_description`/`policy_statement_postfix` — previously broken for every format).
- **Slices instead of maps** for entities/tables, preserving authored document order and ending Go-map-random list order on every load. Source files live on the elements; two side maps deleted.
- **Merge-on-save through the canonical emitters**: parse the existing file, apply only API-editable fields, emit via the excel package's WriteXML — preserving everything the API model doesn't carry, and picking up provenance + number backfill. The old writers rewrote every table from every file through a lossy legacy schema.
- **Numbering semantics**: table lists sorted by TABLE_NUMBER; entity numbers served/persisted; renumber-with-shift (a colliding number shifts everything at/after it down by 100); `POST /api/dt/reorder` + `/api/edd/reorder` for drag-and-drop.
- **`/api/browse`** for the project picker, confined by `--project-root` via the same traversal-safe validation as project open. **`/api/project/current`** so fresh sessions adopt the booted project.
- **`Config.ReadOnly`** rejects all mutations and browsing (reads/compile/execute stay available) for publishing rules for review.

Verified end-to-end against a copy of the staking rules: one save backfilled provenance on all 17 previously provenance-less tables with zero data loss (identical DSL/postfix/initial-action counts). Base: stacked on #908.


PaulSnow:
## Review — feat(apiserver): extract the API server into an importable package

**Overview**: Mechanical move of `cmd/api`'s server into `pkg/dtrules/apiserver` (Config struct replaces package-level flags, `New`/`Routes`/`Run` replace `main`), carrying the session's correctness work: modern-XML parsing (DSL elements, structured initial actions, canonical policy statements), slices-preserving-document-order instead of maps, merge-on-save through the canonical emitters, numbering/reorder endpoints, `/api/browse`, `/api/project/current`, and `Config.ReadOnly`. The #757 parity test moved with the code and passes. **Verdict: merge**, with the caveats below tracked as follow-ups.

**Correctness**
- ✅ Flag conversion is complete — no package-level flag references remain; `--project-root`, CORS origin, and body-size limits all flow through `Config`. The old `-v` flag was declared-but-never-used and is now gone entirely: a behavior-neutral cleanup, but it does remove a documented flag from `cmd/api --help`.
- ✅ Locking survived the split: `LoadProject` takes the write lock internally; `handleProjectOpen` re-acquires a read lock only for the response; `handleProjectCurrent` reads under RLock.
- ✅ Read-only guard composes correctly with CORS (guard inside, CORS outside, OPTIONS allowed through both) and takes an explicit allowlist for the non-mutating POSTs (execute, validate, compile).
- ⚠️ **Known caveat — index-paired merge**: `mergeConditionRows`/`mergeActionRows` preserve postfix and legacy fields by *slice index*. Today's UI can only edit rows in place, so indices always align; but if a future client inserts or removes a row mid-list via PUT, preserved postfix would pair with the wrong row. The postfix would be stale either way (only `dtrules build` recompiles), so the failure is bounded — but a comment upgrade to a hard guard (length mismatch ⇒ drop preservation) would make it unmistakable. Follow-up.
- ⚠️ The deeper authoring-contract gap remains and is inherited, not introduced: the API writes XML without recompiling postfix or writing through to Excel. Merge-on-save makes it non-destructive; contract-compliant saving through the authoring API is the real fix. Deserves a standing issue.

**Conventions/quality**: `Server` methods and route registration read idiomatically; error paths kept the log-and-continue policy that the parity test documents as intentionally different from `cmd/dtrules`. The one map that legitimately remained (`modified`, a set of file paths) is documented by usage.

**Tests**: parity test passes post-move. The new behaviors (shift-on-renumber, reorder, read-only guard, merge-on-save preservation) were verified end-to-end against a copy of the staking rules in-session, but none of that is pinned as Go tests yet — the package is now importable, which makes `httptest`-based coverage cheap. Recommend a follow-up test PR: it would have caught the adoption-vs-read-only interaction found late in UI testing.

PR #910, opened 2026-07-25
feat(editor): sheet-faithful editor UI, embedded in the binary as dtrules edit

PaulSnow:
The UI becomes a working editor that reads like the Excel sheets, and ships inside the dtrules binary (+2.4 MB, 31 MB total): `dtrules edit [project-dir]` serves it at localhost, opens the browser, and boots straight into the project.

**Editor surface**: one grid in the sheet's order (CONTEXTS / INITIAL ACTIONS / CONDITIONS / ACTIONS / POLICY STATEMENTS as banded section rows), rule columns vertically aligned with ruled separators, wrapped comments/DSL, viewing/editing modes, `perform` calls as links with back/forward history, number-edit-with-shift and drag-and-drop reordering, persistent column widths, postfix toggle.

**App shell**: project history + server-side file browser in the Open dialogs; read-only servers strip all editing affordances; Tailwind config restored (the app had shipped without one — every utility class was silently dead); Monaco removed (dead PostfixEditor contradicted the authoring contract; TestPanel uses textarea/pre) making the bundle fully offline.

**Embedding**: `ui/embed.go` behind the `ui` build tag with an npm-free stub; `make build`/`install`/`release` embed by default (`build-noui` opt-out); `dtrules edit` flags `--port/--host/--project-root/--read-only/--no-browser` with a no-auth warning on non-loopback binds; `Dockerfile.editor` (node → go → alpine, non-root, mounted /rules).

Verified: browser-driven tests for navigation, drag-drop, read-only enforcement (server 403s + hidden affordances), and zero-interaction boot on the embedded binary. `make check` green throughout; npm never required for plain Go builds. Docker image build not yet run (no docker perms in the session) — `docker build -f Dockerfile.editor .` wants one verification pass. Base: stacked on the apiserver PR.


PaulSnow:
## Review — feat(editor): sheet-faithful editor UI, embedded as `dtrules edit`

**Overview**: The UI overhaul (single sheet-order grid, view/edit modes, perform-link navigation, numbering + drag-and-drop, project picker/history, read-only awareness) plus the embedding machinery (`ui` build tag with npm-free stub, `dtrules edit` with `--host/--project-root/--read-only`, Makefile defaults, `Dockerfile.editor`). Verified in-session by browser automation on both the dev server and the embedded binary. **Verdict: merge**, with the follow-ups below.

**Correctness**
- ✅ The build-tag split is airtight for CI: only `ui/embed.go` is tag-gated; `edit_cmd.go` compiles untagged against the stub, so `make check`/`go build ./...` never need npm — and a UI-less binary gives a clear actionable message instead of a broken page.
- ✅ `--host` defaults to loopback with an explicit no-auth warning and no browser auto-open on non-loopback binds; `--project-root` reuses the server's traversal-safe validation (verified: `/etc` browse and out-of-root open both rejected).
- ✅ Read-only end-to-end: server 403s are the enforcement; the UI hiding Save/edit/drag affordances is presentation. The GET-only `adoptProject` path exists precisely because the guard (correctly) blocked the old POST-based adoption.
- ⚠️ `http.FileServer` on the embedded FS serves without cache headers; Vite's hashed asset names make this correct but uncached (every page load refetches ~2 MB from localhost — imperceptible locally, mildly wasteful behind a remote proxy). A `Cache-Control: immutable` wrapper for `/assets/` is a cheap follow-up.
- ⚠️ `--port abc` silently keeps 8080 (`fmt.Sscanf` error ignored). Harmless but sloppy; a parse-error exit would be kinder.
- ⚠️ `make check` never compiles with `-tags ui`, so a syntax error inside `ui/embed.go` would only surface at `make build`. One added line in `check` (`go build -tags ui ./ui`) closes that — it needs `ui/dist` to exist though, so gate it or `mkdir -p ui/dist` first. Follow-up judgment call.

**Security posture** (server deployment): no auth by design, documented in help text, Dockerfile, and startup warning — reverse-proxy responsibility is stated everywhere it matters. Read-only mode additionally blocks `/api/browse` so a published server doesn't leak directory listings. This is the right shape for a trusted-network tool; anything internet-facing must front it, and the docs say so.

**Known gaps carried forward** (in-session decisions, restated for the record): the Electron packaging still references the old standalone `cmd/api` layout and wasn't updated; the guided tutorial's steps were only minimally retargeted after the tab consolidation; entity numbers don't yet round-trip through the EDD Excel sheet (flagged on #908); and the Docker image build is unverified in-session (no docker perms) — the layers are individually proven, but run `docker build -f Dockerfile.editor .` once before relying on it.

**Tests**: browser-automation verification was thorough but ephemeral (not committed). The UI has no test harness at all — pre-existing, but this PR triples the behavior surface riding on it. A vitest smoke suite over the store logic (history, adoption, read-only gating, orderAfterMove) would be the highest-value start.

PR #912, opened 2026-07-25
feat(trace): instrument execution tracing end to end; verified replay

PaulSnow:
Brings the trace subsystem up to snuff for the trace debugger. The loader/replayer was built against Java-era fixtures; the Go engine never actually produced usable traces. Full defect list and fixes in the commit message — highlights:

- **Executor instrumented** (`decisiontable` / `execute_table` / `condition` results / fired `column` / nesting `action` elements) — the table/column/action trace methods previously had **zero callers**.
- **`def` events now carry the assigned value's postfix**; previously no value was recorded at all.
- **`intToStr` was broken** (`string(rune('0'+i/1e8))` → `""` for any id < 100M): every array trace event ever emitted had an empty arrayId.
- **New `arraybind` event** ties entity attributes to array ids — without it, replayed arrays floated in a side table no entity referenced (a gap dating to the Java era; the fixtures' tests never verified attached-array contents).
- **Mapping loads now trace** (were explicitly `trace=false`), including the direct `mapping*key` Puts; runtime `addto`/`remove` operators trace, with entity elements recorded by reference.
- **Provenance + final state**: `WriteHeader` records DTRules version + rules fingerprint (sha256 over project XML) per the debugger design; `WriteFinalState` records the full entity stack; `VerifyFinalState` compares replayed state against it attribute-by-attribute.
- **`dtrules run --trace <file>`** produces the complete document; the harness emits the same format.

`TestGoTraceRoundTrip` pins the whole loop on KidAid with real input: execute-with-trace → load → structural assertions → full replay → **zero final-state mismatches**. Java-fixture tests unchanged and passing; `make check` green.

Known follow-ups: `addat`/`removeat` operators still untraced (rare; same helper applies); the harness still doesn't load its test input files (pre-existing, surfaced during this work — separate fix); staking-side trace production goes through the staking repo's SDK wiring.

### DIFF
--- a/cmd/dtrules/run_cmd.go
+++ b/cmd/dtrules/run_cmd.go
@@ -29,6 +29,8 @@ import (
 	"github.com/DTRules/DTRules/pkg/dtrules/interpreter"
 	"github.com/DTRules/DTRules/pkg/dtrules/mapping"
 	"github.com/DTRules/DTRules/pkg/dtrules/session"
+	"github.com/DTRules/DTRules/pkg/dtrules/trace"
+	"github.com/DTRules/DTRules/pkg/dtrules/version"
 	webpkg "github.com/DTRules/DTRules/pkg/dtrules/web"
 )
 
@@ -38,7 +40,7 @@ import (
 // value hasn't been provided (#850/#854).
 func (c *CLI) runRun(args []string) int {
 	path, entry, input, resultEntity := ".", "", "", "result"
-	var save, data, review string
+	var save, data, review, tracePath string
 	interactive, web, noOpen := false, false, false
 	port := "0" // 0 = let the OS pick a free port
 	for i := 0; i < len(args); i++ {
@@ -66,6 +68,11 @@ func (c *CLI) runRun(args []string) int {
 		case "--review":
 			if i+1 < len(args) {
 				review = args[i+1]
+				i++
+			}
+		case "--trace":
+			if i+1 < len(args) {
+				tracePath = args[i+1]
 				i++
 			}
 		case "--result-entity":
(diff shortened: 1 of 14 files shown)


PaulSnow:
## Review — feat(trace): instrument execution tracing end to end

**Overview**: makes the Go engine actually produce traces the Go loader can replay — executor instrumentation in the historical nested format, value-carrying defs, provenance + final state, `dtrules run --trace`, and a KidAid round-trip test that replays to zero mismatches. **Verdict: merge.** Four findings, all fixed on the branch during review:

- **Hot-path allocation**: `traceEscape` built a `strings.Replacer` per event — hoisted to a package var.
- **Trace bloat under iteration**: `arraybind` re-emitted on every push of the same entity; a staking `forall` over 2,847 accounts would have written thousands of duplicate binds. Now emitted once per entity id.
- **`finalState` recorded empty-named attributes** (seen on the primitives entity) — skipped, matching the verifier which already ignored them.
- **Entity-reference attribute writes were untraced** (`data_loader.go` parent-entity linking): now recorded as `refentity`/`refid` defs with a replay handler — the one attribute class a postfix body can't rebuild. The KidAid round-trip passed without this only because no stack entity carried such a reference; the gap was real.

**Remaining follow-ups** (documented, not blocking): `addat`/`removeat` still untraced (same helper applies); the harness reads but never loads test input (pre-existing — worth its own issue since harness "tests" execute dataless); `VerifyFinalState` compares by `StringValue`, so a type change that formats identically would pass — acceptable for the debugger's purpose; staking-side trace production wires up in the staking repo via `SetOutput`/`EnableTrace`/writer.

Affected packages re-tested green; full `make check` green.

PR #1055, opened 2026-08-08
fix(build): --from-excel forces the Excel→XML direction (#1051)

### LINKED ISSUE #1051: build never sets el_compiled, but verify's rebuild does — verify is permanently red for projects maintained with build
## Summary

`dtrules build` writes decision tables back **without** `el_compiled="true"`, while `verify`'s internal Excel→XML rebuild writes them **with** it. The two outputs then differ by exactly that attribute, so `verify` reports `content differs from build output` forever on a project that is otherwise perfectly in sync.

On the Accumulate staking ruleset this is now the **only** remaining difference — every postfix line is byte-identical:

```
$ dtrules verify --diff
[build] xml/staking_dt.xml: content differs from build output
-<decision_table>
+<decision_table el_compiled="true">
```

Two differing lines, 35 tables, nothing else.

## Cause

`pkg/dtrules/excel/dt_importer.go:655` emits the attribute from the in-memory flag:

```go
if table.ELCompiled {
    f.WriteString("<decision_table el_compiled=\"true\">\n")
} else {
    f.WriteString("<decision_table>\n")
}
```

When verify imports into an empty temp directory the tables are freshly compiled from EL, so the flag is true. When `build` runs in place, the flag is read back from the existing XML — false for any project whose XML predates the attribute — and written out false again. The postfix **is** recompiled from EL in both cases; only the marker disagrees.

So a project can never reach a state where `build` output and `verify` output match, short of hand-editing the attribute in — which the authoring contract forbids.

## Why it matters

`verify` is the gate the authoring contract leans on. A gate that cannot go green on a correct project is one people learn to ignore, and it hides real findings behind a permanent known-failure. We have had to leave `verify` out of our `verify-all` target for exactly this reason, which is the outcome the contract is trying to prevent.

## Suggested fix

Set `ELCompiled = true` whenever the postfix is generated from EL, regardless of what the previous XML said — the flag describes the write that is happening, not the file being replaced. Alternatively, treat the attribute as metadata and exclude it from verify's comparison; but the first is more honest, since after a build the postfix genuinely *was* EL-compiled.

## Context

Found finishing the migration of a production ruleset onto the Excel-as-record contract. The related fixes (#993, #1006, #1010, #1029, #1031) have landed and this ruleset now round-trips with byte-identical postfix — this attribute is the last thing standing between it and a green gate.

PaulSnow:
Closes #1051.

`--from-excel` selected the Excel-authored code path but did not set a direction. `SyncAll` still ran its per-workbook timestamp detection and answered `NoSync` whenever the XML was not older, so the import never happened:

```
XML is already up to date.
tables=0  actions=0  conditions=0  entities=0  mappings=0
files-written=0
```

The flag asked for a direction and got a timestamp comparison.

### The reported symptom (#1051)

A table could never acquire `el_compiled` through `build`. The attribute is written from the in-memory flag, which is set when the postfix is compiled from EL — and with the import skipped, nothing is compiled. `verify` rebuilds into an empty directory, which is always written, so its output always carried the attribute. `verify` therefore reported `content differs from build output` **permanently**, on a project that was otherwise byte-identical.

### The quieter and worse symptom

For a project where Excel is the system of record, an edit made in Excel is **silently discarded** whenever the XML happens not to be older — with a zero exit and a line stating the XML is already up to date. There is no signal that the edit was dropped.

### Fix

```go
syncOpts.ForceDirection = dtrsync.ExcelToXML
```

`ForceDirection` already exists; `--from-excel` simply never used it.

### Verification

Against the Accumulate staking ruleset (35 tables), built from this branch vs. committed `main`:

| | `main` | this branch |
|---|---|---|
| `build --from-excel` | `tables=0 … files-written=0` | `tables=35 actions=101 conditions=56 entities=18 files-written=2` |
| rebuilt XML | (unchanged — nothing ran) | **byte-identical** to committed |
| `el_compiled` | 35 | 35 |
| `fphalfup/` divisions | 10 | 10 |
| `verify` | pass | pass |

So the import now actually runs and reproduces the ruleset exactly — including every round-half-up division, which is the money path.

### DIFF
--- a/cmd/dtrules/build.go
+++ b/cmd/dtrules/build.go
@@ -433,6 +433,21 @@ func (c *CLI) runExcelAuthoredBuild(xmlDir, excelDir string, opts *buildOptions)
 
 	syncOpts := dtrsync.DefaultOptions()
 	syncOpts.Verbose = opts.verbose
+	// --from-excel means Excel-authored, so say so instead of leaving the
+	// per-workbook timestamp check to decide. Selecting this branch only
+	// chose the code path; SyncAll still detected direction per workbook
+	// and answered NoSync whenever the XML was not older, so the import
+	// never ran: "XML is already up to date", tables=0, files-written=0.
+	//
+	// The visible cost was that a table could never acquire el_compiled.
+	// The attribute is written from the in-memory flag, which is set when
+	// the postfix is compiled from EL — and with the import skipped
+	// nothing was compiled, so an XML predating the attribute could not
+	// gain it through build, while verify's rebuild (into an empty
+	// directory, always written) always had it. verify then reported
+	// `content differs from build output` forever, on a project that was
+	// otherwise byte-identical (#1051).
+	syncOpts.ForceDirection = dtrsync.ExcelToXML
 
 	syncer := dtrsync.NewSyncerWithOptions(xmlDir, excelDir, syncOpts)
 	syncer.SetUseCombinedWorkbooks(true)


PaulSnow:
Closing as superseded: main already carries this fix, with tests.

PR #1057 (`d9d6d64a`, merged 2026-08-08 16:45) landed the same `syncOpts.ForceDirection = dtrsync.ExcelToXML` one line — about an hour after this branch was authored (15:44) — plus `cmd/dtrules/build_from_excel_test.go`, whose `TestBuildFromExcelImportsEvenWhenXMLIsNotOlder` pins exactly the behavior this PR describes: XML no older than Excel, import runs anyway. #1051 is closed by it. Both tests pass on current main.

Reviewed this branch before closing rather than instead of it:

- **The fix is identical** — same field, same value, same function. Merging main into this branch produces two consecutive `ForceDirection = ExcelToXML` assignments, which is the clearest possible evidence of duplication.
- **#1057's version is a strict superset**: its comment covers all three ways execution reaches that branch (`--from-excel`, no `xml/` yet, or detection saying Excel is newer), where this one explains only `--from-excel`; and it brought the regression test this PR has none of.
- **The failing `verify` check was unrelated** — KidAid sample drift as of Aug 8 (`xml/kidaid_dt.xml`/`_edd.xml` differing from build output), since repaired: KidAid verifies clean on main today.

Nothing here is lost by closing. Two implementations of one issue within the hour is the coordination hazard of parallel sessions — the same thing that produced #1111/#1120 — not a defect in either branch.

Issue #1020, opened 2026-08-08
EL compiles unknown and misspelled operator statements at any arity; the failure waits until the row executes


PaulSnow:
Found while evaluating a full conversion of `github.com/paulsnow/cribbage` onto the #980 primitives.

Every combinatorial primitive is invoked as a statement-form operator call (`typedOperator : IDENT`, as #980 designed). Nothing on that path checks the name against the operator registry, or the argument count against the operator. Probed against `CompileAction` (pkg/dtrules/compiler/el):

```
subsets(hand.cards, "combo", "value", hand.combos)   OK -> hand.cards "combo" "value" hand.combos subsets
subests(hand.cards, "combo", "value", hand.combos)   OK -> hand.cards "combo" "value" hand.combos subests   <- typo
frobnicate(hand.cards, "x", 3, hand.combos)          OK -> hand.cards "x" 3 hand.combos frobnicate          <- invented
subsets(hand.cards)                                  OK -> hand.cards subsets                              <- 1 of 4 args
windows(stack.cards, 3, "combo", peg.tails)          OK -> stack.cards 3 "combo" peg.tails windows          <- unimplemented
```

None of `subests`, `frobnicate`, `windows` is registered (`grep Register\( pkg/dtrules/operators/`). It survives postfix compilation too, and the failure waits for execution:

```
CompilePostfix(`hand.cards "combo" "value" hand.combos subests`) -> err=<nil>
execute -> [Undefined] RName: The Name 'subests' was not defined by any Entity on the Entity Stack
           PostFix: hand.cards "combo" "value" hand.combos  ERROR==> subests  <==
```

Two problems:

- **Nothing catches it at authoring time.** A misspelled primitive in a column that rarely fires compiles, saves, and ships; it fails the first time that column fires in production. The `subsets(hand.cards)` case is worse — the operator name is real and the arity is silently wrong.
- **The runtime error points at the wrong thing.** "not defined by any Entity on the Entity Stack" reads as a data/EDD problem. The reader looks for a missing field; the bug is a typo in an operator name.

This hazard grows with the primitive family: each of `subsets`/`combinations`/`groupby`/`maximalruns` is a bare IDENT taking 3-5 positional arguments, several of which are EDD type and field names passed as strings — no compile-time contract on any of it.

## To do

- Resolve statement-form operator names against the registry at compile time. Unknown name = compile error naming the table, the row, the operator, and ideally the nearest registered name.
[...]

PR #1111, opened 2026-08-12
fix(el): reject operator calls with the wrong argument count (#1105)

### LINKED ISSUE #1105: Operator arity is still unchecked — the deferred half of #1020
Deferred half of #1020. That fix (#1046) resolves statement-form operator *names* against the registry, and its own commit message says so plainly:

> Arity is not checked. The registry records no argument count, so `subsets(hand.cards)` with one of four arguments still compiles; that needs arity metadata and is a separate piece of work.

No issue was opened for it, so here it is. Re-verified on v1.23.1 (`6af41faf`):

```
subsets(hand.cards)                                     OK -> hand.cards subsets
subsets(hand.cards, "combo", "value", hand.combos, 7)   OK -> hand.cards "combo" "value" hand.combos 7 subsets
```

Both compile clean. The four-argument form is the only correct one.

## Why the remaining half still bites

The name check catches typos, which was the loud failure. Arity is the quiet one, and for these operators it is worse than a wrong name:

- **A short call leaves the stack wrong, not empty.** `hand.cards subsets` pops whatever three values happen to be beneath it and treats them as typename/sumfield/dest. Depending on what the row did before, that is a runtime error, a write into the wrong array, or — worst case — a plausible-looking result.
- **The #980 generators take 4-5 positional arguments, three of them bare strings** naming EDD types and fields. Nothing about `subsets(hand.cards, "combo", "value", hand.combos)` reads wrong if an argument is missing; the shape is not memorable enough for a reviewer to catch it in a spreadsheet cell.
- It is the same authoring-time-vs-runtime asymmetry #1020 was filed about, and the same one #1046's commit message calls "silent acceptance, late failure."

## To do

- Record arity in the registry alongside the name — minimum, and whether variadic. `Register` is the single choke point (`pkg/dtrules/operators/registry.go`), and `registration_test.go` already keeps a hard-coded list of every operator name that would have to be extended in step, which gives the change a natural test.
- Check it in the same injected hook #1046 added (`SetOperatorChecker` in `pkg/dtrules/compiler/el/compiler.go`, injected from `pkg/dtrules/authoring/el_check.go` and `cmd/dtrules/build.go`) — probably by widening the callback from `func(name string) bool` to something that can also answer "how many arguments".
- Error message should name the operator, the count given, and the count expected.
- Where cheap, the string arguments naming EDD types and fields are checkable against the loaded EDD too — a separate step, but the same hook.

## Acceptance

`subsets(hand.cards)` fails `dtrules build` naming the table, the row, and the expected argument count; all correct calls in the Cribbage sample's eleven tables still compile; `make check` green.

PaulSnow:
Closes #1105 — the half #1020 deferred.

#1046 resolved statement-form operator names against the registry and said in its own commit message that arity was left for later: the registry recorded no argument count, so `subsets(hand.cards)` with one of four arguments still compiled clean.

The name check catches typos, the loud failure. Arity is the quiet one, and for these operators it is the worse of the two: the runtime pops arguments by position, so a short call does not fail for want of arguments — it takes whatever values sit beneath it on the stack and treats them as the type name, sum field and destination array. Depending on what the row did before, that is a runtime error, a write into an unrelated array, or a plausible-looking wrong answer.

## Approach

- Arity is declared next to `Register`, and **deliberately sparse**. Most operators are reached through EL syntax the grammar already shapes (`sum of x in y`), where the count cannot be miscounted; only the `name(a, b, …)` form can be. An operator with no entry is not checked, so this does not become a table that must stay exhaustive to be correct.
- Injected rather than imported, as in #1046, and set on the same two authoring paths: the authoring SDK and `dtrules build`.
- Counting walks the right-recursive `operatorlist` chain looking for a nested operatorlist among the children, rather than type-switching over all eight labelled alternatives — which would need revisiting every time the grammar grows an argument type.

## Verification

- Six wrong-count calls refused, each error naming the expected count.
- The five correct forms still compile; an operator with no declared arity is untouched.
- The Cribbage sample's eleven tables still build from their workbook and `dtrules verify` passes.
- `make check` green.

### DIFF
--- a/cmd/dtrules/build.go
+++ b/cmd/dtrules/build.go
@@ -48,6 +48,11 @@ func newWorkbookImporter(xmlDir string) *excel.WorkbookImporter {
 		_, ok := operators.GetByString(name)
 		return ok
 	})
+	// And with the declared number of arguments (#1105).
+	c.SetOperatorArity(func(name string) (int, int, bool) {
+		a, ok := operators.ArityOf(name)
+		return a.Min, a.Max, ok
+	})
 	imp.SetELCompiler(c)
 	if syms := authoring.LoadEDDSymbols(xmlDir); len(syms) > 0 {
 		imp.SetSymbols(syms)
--- a/pkg/dtrules/authoring/el_check.go
+++ b/pkg/dtrules/authoring/el_check.go
@@ -136,5 +136,12 @@ func newCheckedCompiler() *el.Compiler {
 		_, ok := operators.GetByString(name)
 		return ok
 	})
+	// And with the right number of arguments — a short call is worse than a
+	// misspelled one, because it silently consumes whatever is beneath it on
+	// the stack rather than failing (#1105).
+	c.SetOperatorArity(func(name string) (int, int, bool) {
+		a, ok := operators.ArityOf(name)
+		return a.Min, a.Max, ok
+	})
 	return c
 }
--- a/pkg/dtrules/authoring/operator_arity_test.go
+++ b/pkg/dtrules/authoring/operator_arity_test.go
@@ -1,0 +1,93 @@
+// Copyright 2026 Paul Snow
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authoring
+
+import (
+	"strings"
+	"testing"
+)
+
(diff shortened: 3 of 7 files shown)


PaulSnow:
Closing as superseded: main already carries this fix. PR #1120 (commit 4651d943) landed the same #1105 arity check while this branch sat — `RegisterWithArity` on the operator itself, all five combinatorial operators covered with the same counts this branch declares, `-1` documented as "do not check", and its own authoring-level tests. Issue #1105 is closed by it.

Reviewed this branch before closing rather than instead of it. The designs are equivalent where they overlap; the two real differences, for the record:

- This branch's `(min, max, ok)` lookup supports **ranges and variadic arity**; #1120's single `int` cannot. Nothing registered today needs it — all five ops are fixed-arity — so this is the capability to resurrect if a variadic statement-form operator ever appears, not a reason to churn a landed design.
- This branch's sparse side-map vs #1120's field-on-Operator: same semantics, different storage; no reason to prefer either after the fact.

Two implementations of one issue in three days is the coordination hazard of parallel sessions, not a defect in either. The branch is preserved in the PR record if the range capability is ever wanted.

Issue #1025, opened 2026-08-08
Table-execution cost (~45 µs, 39 KB per five-card scoring) is the ceiling on enumerate-and-average (#980's `expect`)


PaulSnow:
Measured while evaluating a conversion of `github.com/paulsnow/cribbage` onto the #984 tables, embedding DTRules in Go with the host owning input/output and display.

**The tables are exact.** 3,000 random deals scored by `Score_Hand` against the reference `scoreHand`, comparing all five categories and not just the total: **0 mismatches**. Fifteens from `subsets`, pairs from `groupby`, runs from `maximalruns`, the four-vs-five-card flush rule, nobs — all reproduce the reference implementation.

**The cost, per scoring of one five-card hand:**

| | |
|---|---|
| DTRules `Score_Hand` | **~45 µs** |
| Go reference `scoreHand` | **~186 ns** |
| allocation | **~39 KB** |

~250x, and **94% of it is inside `dt.Execute`** — entity creation and array setup on the host side account for only ~3 µs. So this is not the embedding boundary; it is the table execution itself.

## Why this matters beyond one card game

The trainer's actual product is *exact expected value*: every one of 15 discards scored against every starter and every opponent toss — 15 x (46 + 45,540) = **683,790 scorings per deal**. In Go that is 127 ms single-threaded, well under a second across 15 goroutines. At table speed it is **~33 s single-threaded** and ~27 GB of allocation churn. Sessions are not thread-safe (docs/architecture.md:532), so parallelism costs one session per worker — cheap at 27 µs each, and not the problem.

That gap is exactly what makes #980's *other* later candidate — "`expect`: enumerate-and-average over deals for EV analysis" — not yet worth building. A native enumerator still pays the per-execution cost on every inner call, so it turns 33 s into maybe 25 s, not 0.3 s. **The enumeration is not the cost. The table call is.** Worth settling before anyone designs `expect`.

To be clear about scope: the EV kernel is not policy and belongs in Go — no analyst will ever edit "average over 46 starters". But an order of magnitude here moves the boundary of what can be *authored* instead of *coded*, which is the point of the engine.

## What to look at

- Per-execution overhead: entity stack push/pop, context frames, the work arrays the EDD requires per run.
- `subsets` on 5 cards materializes 31 combo entities plus 31 member arrays on every call, all garbage a moment later. A reusable destination array or an entity pool across executions is the obvious lever, and it applies to all four #980 generators.
[...]

PR #1132, opened 2026-08-15
perf(runtime): the scratch arena — recycle what the generators materialize (#1025)

PaulSnow:
The arena decided on #1025, built to the recorded design:

- **Opt-in** (`EnableScratch` / `ResetScratch`): zero change and zero pinned memory for hosts that never opt in.
- **The escape hatch fails loudly**: generation-stamped scratch entities error at Get/Put after a reset — never a silent read of recycled storage. One nil-check on ordinary entities.
- **Fresh IDs every reuse** — no two incarnations behind one trace ID.
- **Reuse = re-init**: `CloneEntity`'s value initialization is now the shared `reinitFrom`; a recycled instance is a fresh clone minus the allocations.

Profiling the arena also surfaced an unconditional engine cost: **trace-attribute `fmt.Sprintf` on every condition/action evaluation with tracing off (~8% CPU)**. `State` gains `Tracing()`; the node hot paths guard their formatting. That fix helps every execution.

## Numbers (benchtime 5000x, ×3 runs)

| | ns/op | B/op | allocs |
|---|---:|---:|---:|
| main before this arc | 54,810 | 42,104 | 791 |
| after #1128 | 49,166 | 40,878 | 748 |
| this PR, no arena | ~45,500 | 36,372 | 592 |
| this PR, arena | **~38,300** | **17,694** | **428** |

## The honest correction

`docs/table-execution-cost.md` now records it: materialization is ~⅓ of execution, not ~95% — the profile puts `emitCombo` at ~33% cumulative. Bytes halve under the arena (the allocation story *was* materialization-shaped); time improves ~16% because CPU mostly wasn't. The next order of magnitude, if ever needed, is interpreter-loop work.

Tests: arena unit suite (pointer-identity recycling, fresh IDs, re-defaulted values, stale-use errors, array recycling, disabled fallback) plus an integration gate alternating the perfect-29 and double-run hands across six resets. `make check` green. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/table-execution-cost.md
+++ b/docs/table-execution-cost.md
@@ -71,6 +71,37 @@ no analyst will ever edit "average over 46 starters", and it belongs in Go at
 190 ns per scoring. The point of measuring is to know where that boundary
 actually falls, and to notice if it moves.
 
+## The arena, and what it actually recovered (#1025 follow-up)
+
+The scratch arena landed with these revisions to the numbers above. Same
+machine, `-benchtime 5000x`:
+
+| Benchmark | ns/op | B/op | allocs/op |
+|---|---:|---:|---:|
+| `BenchmarkCribbageScoreHand` (no arena) | ~45,500 | 36,372 | 592 |
+| `BenchmarkCribbageScoreHandArena` | ~38,300 | 17,694 | 428 |
+
+Two things moved everything since the first table:
+
+1. **Trace formatting was ~8% of execution with tracing off.** The
+   condition/action nodes formatted `fmt.Sprintf` trace attributes on every
+   evaluation; the Trace* methods no-op internally, but Go evaluates
+   arguments regardless. Guarded behind `State.Tracing()` — this one helps
+   every execution, arena or not.
+2. **The arena recycles what the generators materialize.** Opt in with
+   `sess.EnableScratch()`, call `sess.ResetScratch()` between executions;
+   combo/group/run entities and their members arrays are recycled with
+   fresh IDs, and any reference held across a reset fails loudly at next
+   use instead of reading recycled storage.
+
+And one correction to the earlier analysis: materialization is **about a
+third** of execution, not ~95% — the profile puts `emitCombo` at ~33%
+cumulative, with the rest in bytecode dispatch, entity-stack machinery, and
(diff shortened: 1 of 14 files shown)


PaulSnow:
Merging over the red check, with the reasoning on record: the failure is #1133 — the #870 runner kill returned on **main itself** at 5d3295b1, two PRs before this branch existed, killing every run ~130-150s in with no test failure. This branch passes the full `make check` in both modes locally (twice each), and the `authoring` binary's peak RSS is identical on this branch and main (~330 MB), so the merge neither causes nor worsens the condition. Four CI attempts all died with the same signature main dies with.

PaulSnow:
Fixed in #1136.

The export's rule-set load no longer builds decision trees. Validation and expression compilation still run, so an export still refuses a table it cannot compile; only the tree and the unreachable-column analysis that reads it are skipped.

| | before | after |
|---|---|---|
| `TestSave_TaxReturnIdempotent` peak RSS | 29.3 GB | **329 MB** |
| wall clock | 70 s | **0.24 s** |
| whole `authoring` package | OOM-killed | 3 s, 330 MB |

All 12 samples still verify with zero findings, and an authoring write still produces correct Excel and XML.

Leaving the wider point on the record rather than in this fix: the eager build remains for every other consumer. `verify`, `review` and the analysis passes all load rule sets they never execute, and each is paying some version of this. Making the tree lazy — built on first execute — would fix the class rather than one call site, but it moves `setUnreachable` off the load path, so the advisory output has to be checked before doing it. Worth its own issue if the memory shows up elsewhere.

PaulSnow:
The "worth a second look" from this issue is done: #1149 made trees build on first execution (fixing every loading consumer — verify, review, analysis — not just the export path), and #1150 fixed the tree cost itself via FIRST-capable lazy encoding. One TaxReturn load: 24.5 GB → 13 MB. The archive lane: ~400 s and OOM-prone → 3.8 s at 273 MB.

PaulSnow:
Correction: two comments and several commit messages (#1136, #1149, #1150) cite #1132 for the decision-tree memory work. That was a transposition — the memory issue is **#1135**. This PR (the scratch arena) is unrelated to those references. Sorry for the noise on your thread.

Issue #1091, opened 2026-08-10
Excel-hash provenance in the XML: retire .sync-manifest.json and every mtime comparison


PaulSnow:
Design agreed with Paul. Recording it with the evidence, because the two
preconditions are now in and this is ready to build.

## The idea

Excel is the system of record, so the relationship is one-directional and one
hash is enough. Two files cannot hold each other's hashes; only the derived
artifact records where it came from.

```
authoring edits the model
  → writes Excel                     (system of record)
  → compiles Excel → XML             (derived)
  → stamps the XML with hash(Excel)  (provenance)
```

The check is two file reads:

```
hash(workbook on disk) == <source_hash> recorded in the XML ?
```

Not equal means the XML is stale — rebuild it. No manifest, no rebuild, no
timestamps.

The XML already has the place to put it: every table carries a `<source>`
block naming its workbook. It gains one element beside `<file_name>`.

## Why not what we have now

Timestamps answer "which was touched last", which is a proxy for "which
changed", and the proxy is wrong constantly — checkout, `cp`, containers, CI,
mtime granularity. Most of this month's sync bugs were that proxy misfiring:

- **#1057** — `build --from-excel` asked mtimes, got NoSync, imported nothing, exited 0
- **#1061** — the guard compared mtime to a recorded export time, and checkout stamps every file, so **every fresh clone started locked**
- **#1069** — `sync export` let mtimes overrule the direction the command's own name stated

`.sync-manifest.json` is also gitignored, so the baseline never travels. That
is the root of #1061 and it cannot be fixed while the state lives outside the
artifacts. Provenance in the XML travels with it through git, `cp` and CI.

## Preconditions, both now met

**The round trip is lossless.** XML → Excel → XML is byte-identical on
SinusitisTherapy, SyntaxTests, CHIP, Poker, Cribbage and StateTax. That only
became true with #1085's dense rows; before it the round trip added
`column_value="-"` entries. Making the XML fully derived costs nothing.

**An export that cannot carry everything refuses** (#1088). Under this model a
silently-skipped table is not an incomplete workbook, it is a deleted rule —
the next build regenerates the XML from a workbook that never received it.

## What it does not cover

Provenance detects "someone changed one side without regenerating". It does not
detect "the generator itself is wrong", and that is the failure class that hurt
most this month:

- **#1086** — every TaxReturn workbook holding all 205 sheets. The refresh wrote
  those workbooks and would have written matching hashes. Provenance agrees;
[...]

PR #1153, opened 2026-08-19
feat(sync): the manifest is retired by attrition (#1091)

PaulSnow:
Nothing reads `.sync-manifest.json` state any more — direction and the export
guard ask the provenance hash in the XML (#1124), and workbook pairing comes
from the artifacts' own source blocks (#1130). What remained: every export
still wrote the file, and every new project got a `.gitignore` whose only
content was hiding it.

## Change

`RecordExport` refreshes a manifest **only where one already exists**: a
project that has one never disagrees with itself; a project without one never
grows one. The `.gitignore` write is deleted — gitignoring the manifest is the
exact reason a fresh clone used to arrive with no sync state at all, which was
this issue's founding complaint.

## Measured

Fresh `git archive` of SinusitisTherapy → `table put` → `build --from-excel` →
`verify`: **zero manifests created**, verify clean, zero findings.

Reading stays (`LoadManifest`, the fallback pairing) for XML predating source
blocks; deleting the reader is for after the last committed manifest leaves the
samples.

`make check` passes.

### DIFF
--- a/pkg/dtrules/sync/manifest.go
+++ b/pkg/dtrules/sync/manifest.go
@@ -197,8 +197,16 @@ func (m *Manifest) CheckExcelModified(excelPath string) (modified bool, entry *F
 }
 
 // RecordExport records that XML was exported to an Excel file.
+//
+// Since #1124/#1130 the artifacts carry their own provenance and pairing, so
+// nothing reads this state any more. A project that already has a manifest
+// keeps it current so it never disagrees with itself; a project without one
+// never grows one -- the file is retired by attrition (#1091).
 // Call this AFTER a successful export to update the manifest.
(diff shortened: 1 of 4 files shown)


PaulSnow:
Closing without merging — merge-pr.sh correctly refused it. The RecordExport-level gate breaks eleven tests across three packages: NewManifest()+RecordExport is a legitimate in-memory workflow (path unset until Save), and the enforcement-guard tests build their baseline through exactly the creation path I gated. The attrition idea is right; the gate belongs at the project-facing creation sites (the sync export flows around sync.go:868/1422/1492), not inside RecordExport. Handed off on #1091 with specifics.

Issue #1168, opened 2026-08-30
dtrules run --input silently discards data that lands on a singleton entity


PaulSnow:
`dtrules run --input` returns EDD defaults instead of the scenario's data whenever a mapped attribute's enclosure is an entity pushed by `<initialentity>`. Nothing errors; the run completes and prints a plausible, wrong answer.

## Repro

```bash
dtrules run ./sampleprojects/TaxReturn --entry Compute_Tax_Return \
  --input sampleprojects/TaxReturn/testfiles/TestScenarios/Integration/TestCase_Integration_01_Family_All_Credits.xml \
  --result-entity job
```

The scenario says `<state>OH</state>`, `<filing_status>MFJ</filing_status>`, `<expected_agi>73200</expected_agi>`. The run reports:

```
  filing_status:
  state:                 TX      <-- the EDD default, not the scenario
  expected_agi:
```

Total tax computes as $0 and the audit trail says `State Tax Calculation for TX`. Passing `--map sampleprojects/TaxReturn/xml/TaxReturn_map.xml` explicitly makes no difference.

The same file, same mapping, loaded the way `pkg/dtrules/scenario_coverage_test.go` does it — `LoadMapping` → `Initialize` → `LoadData` — gives `job.state="OH"`, `filing_status="MFJ"`, `expected_agi="73200"`, and validation runs. So the data and the mapping are fine; the CLI path is not.

## Cause

`runCmd`'s loader (`cmd/dtrules/run_cmd.go:289-299`) calls `m.LoadDataAndPushSingletons(f)` instead of `Initialize()` + `LoadData()`. Its own comment states the intent:

```go
// Load the data first, then push singletons so the stack holds the
// LOADED instances — Initialize+LoadData would push default-valued
// entities disconnected from the input.
```

That is the opposite of what happens. TaxReturn's mapping declares `<initialentity entity='job' epush='true'>` (`TaxReturn_map.xml:268`), and the `enclosure='job'` setattributes resolve against the entity on the stack (`pkg/dtrules/mapping/data_loader.go:177-206`). With the singleton pushed *after* the load, those attributes have no `job` to land on, and the freshly pushed default-valued `job` is what the run then executes against.

CHIP is unaffected — its data arrives through `createentity` lists rather than attributes on a pushed singleton — which is presumably why this survived.

## Impact

[...]

Issue #1169, opened 2026-08-30
Authoring writes into xml/states/* skip Excel silently and break dtrules verify


PaulSnow:
An authoring write to a table in a subdirectory reports success, does not update the paired workbook, and leaves the project failing `dtrules verify`. The authoring contract's third invariant says the Excel write-through is "mandatory and **fail-closed**: an API write that cannot land Excel must fail, not silently skip it". Here it silently skips.

## Repro

On a copy of `sampleprojects/TaxReturn` (verify passes at baseline):

```bash
echo '{"op":"update-condition-dsl","condition_number":1,"dsl":"job.state is equal to \"CO\""}' \
  | dtrules table patch CO_Tax --project <copy>
```

```
{ "file": "states/CO_dt.xml", "op": "update-condition-dsl", "status": "patched", ... }
```

`excel/states/CO.xlsx` is byte-identical afterwards (md5 unchanged). Then:

```
$ dtrules verify <copy>
[build] xml/states/CO_dt.xml: content differs from build output
exit 1
```

The XML moved, the workbook did not, so the committed XML no longer matches what a build from Excel would produce. The next `dtrules build` overwrites the edit.

## Cause

`discoverWorkbookPairing` (`pkg/dtrules/authoring/workbook_pairing.go:44-63`):

```go
abs := filepath.Join(excelDir, name)          // excel/CO.xlsx — does not exist
if _, err := os.Stat(abs); err != nil {
    if alt := filepath.Join(excelDir, filepath.Base(name)); alt != abs {  // same path; never taken
```

State tables record `<file_name>CO.xlsx</file_name>` — a bare basename — which resolves to `excel/CO.xlsx`. The real workbook is `excel/states/CO.xlsx`. The `filepath.Base` fallback is a no-op for a name that is already a basename, so the pairing is never found, and `RefreshExcelIn` (`pkg/dtrules/authoring/excel_sync.go:227`) hits the failing `os.Stat` and `continue`s without exporting or erroring.

`.sync-manifest.json` does key it correctly as `states/CO.xlsx`, but that branch only runs when the discovered pairing is empty, and it isn't.

Affects `table put`, `table patch`, and `edd patch` against any `xml/states/*` file — the whole per-state layout `.claude/CLAUDE.md` recommends for avoiding merge conflicts.

## Suggested fix

[...]

Issue #1047, opened 2026-08-08
Rebuilding CHIP shifts a local-variable slot and the result fails with GetFrameValue OutOfBounds


PaulSnow:
Found while sweeping the sample drift in #1012, which I had expected to be mechanical after this week's round-trip fixes. It is not: for CHIP a from-Excel rebuild silently changes what executes.

## Evidence

Rebuilding `sampleprojects/CHIP` from its own workbook changes **8 of 65** postfix elements. The build reports `drops: none` and exits 0.

Two action rows lose their postfix entirely:

```
DSL : add "The System does not yet support Medicaid Applications" to the client.notes
 was: "The System does not yet support Medicaid Applications" client.notes swap addto
 now: (EMPTY)
```

The DSL is present in the workbook (checked the shared strings) and compiles correctly on its own:

```
CheckAction(...) -> "The System does not yet support Medicaid Applications" client.notes swap addto
```

So the row survives the export, and the compiler handles it — it is the import that drops the result.

A third row shifts a local-variable slot:

```
DSL : set the client_fpl = fpl
 was: 0 local@ cvi /client_fpl xdef
 now: 1 local@ cvi /client_fpl xdef
```

That one is not cosmetic. The rebuilt rules fail at execution:

```
[OutOfBounds] GetFrameValue:
PostFix: /Evaluate_CHIP_Eligibility  ERROR==> performtable  <==
```

`TestCHIPPerformance`, `TestCHIPSimpleExecution`, `TestCHIPCountsTheHousehold`, `TestKidAid` and the trace-floor test all fail against the rebuilt tables. The committed XML passes.

## Why this matters more than the drift it was found in

#1012 is about `verify` reporting a difference. This is worse: taking the advice implied by that difference — rebuild from Excel, the system of record — produces rules that do not run. A project cannot be brought into contract compliance by the documented route.

It also means the remaining `[build]` findings on CHIP, ChipApp, KidAid and TaxReturn cannot be cleared by a build sweep, which is what I had assumed. Poker and SinusitisTherapy rebuild with zero DSL or postfix change and are safe.

## What to look at

The local-slot shift suggests context/local allocation differs between the committed compile and a fresh one — possibly an ordering dependency in how locals are assigned across a table's sections. The two emptied rows are a separate question: DSL present, compiles standalone, postfix empty after import.

## Reproduction

```
cp -r sampleprojects/CHIP /tmp/chk && rm -f /tmp/chk/excel/.dtrules-manifest.json
[...]

Issue #965, opened 2026-07-28
Locals declared in a context row are invisible to the table's condition and action rows


PaulSnow:
Blocks CHIP (#962) and, by inspection, ChipApp, KidAid and SyntaxTests — four of the seven samples left in campaign #948.

## The shape

CHIP's `Calculate_Group_Size` declares a local in its context block and refers to it from every row:

```
context:    for all clients whose applying == true
context:    local entity ApplyingClient = client
context:    for all clients
condition 1: ApplyingClient == client
action 1:    add 1 to ApplyingClient's IncomeGroupCount
```

The context compiles fine — `client cve allocate execute deallocate pop` allocates the slot. The rows do not:

```
condition 1: ApplyingClient == client  ->  ApplyingClient client req
```

`ApplyingClient` is emitted as a bare name, and at run time:

```
[Undefined] RName: The Name 'ApplyingClient' was not defined by any Entity on the Entity Stack
```

## Why

Every row is compiled as its own unit against the EDD symbol table alone — `compileDSLOrEmpty(dsl, t.symbols, kind)` in `authoring/table.go`. The emitter has `lookupLocal` / `emitLocalRef`, so locals work *within* one compilation unit, but nothing carries the slots a context declared into the compilation of the rows beneath it.

So the idiom parses, compiles, produces plausible postfix, and fails at execute. It has presumably never worked: CHIP's checked-in postfix used an older name (`ThisClient`) that fails the same way, so this table has been dead for as long as the sample has existed.

## Why it matters beyond CHIP

`grep -l "local entity" sampleprojects/*/xml/*.xml` hits CHIP, ChipApp, KidAid and SyntaxTests. Any repair of those four has to either make this work or re-author around it.

## Options

1. **Compile a table's rows with its context locals in scope.** Correct, and what the EL implies. Needs the compile entry points to take the table's declared locals, which changes the authoring API's per-row compilation model.
2. **Re-author the samples around it** — hold the value in a field (`set job.applying_client = client`) rather than a local. Works today; changes four samples and loses the idiom.

Worth a decision before the CHIP/ChipApp/KidAid/SyntaxTests repairs, since it determines whether those are engine work or authoring work.

PR #1171, opened 2026-08-30
feat(TaxReturn): reciprocal state agreements (#234)

PaulSnow:
Implements #234. **Stacked on #1170** (`feat/map-create-entity-list-1164`) — review/merge that first; this PR's base will retarget to `main` once it lands.

A resident of state A working in state B pays only A: B does not tax the wages, and on a filed certificate does not withhold on them. 30 bilateral agreements across 16 jurisdictions, each verified against the work state's own revenue department publication.

## The issue's premise was false

#234 asks to "skip non-resident tax when reciprocity applies". There was nothing to skip:

| Piece | State before this PR |
|---|---|
| `state_period` mapping | absent from `TaxReturn_map.xml` — no `createentity`, no `setattribute` |
| `job.state_periods` at runtime | **0**, on every scenario |
| `Calculate_State_Source_Income` (40500) | orphaned — no `perform` reaches it |
| `is_resident`, `state_source_income` | written by table 40510, **not declared in the EDD** |
| per-state wage income | did not exist; the allocator pooled every income type into one double |
| `income.state_code` | also unmapped, so the allocator could never match income to a period |

`Dispatch_State_Tax` branches on the scalar `job.state`, so only the resident state is ever computed. Reciprocity applied to that would have been a no-op. #231/#232/#233/#235 are all closed but their machinery was never connected. So this wires the path up and then applies reciprocity to it.

## Changes

- **map** `state_period` (and `income.state_code`), so multi-state input stops being dropped at load
- **split wages from other income per period.** Reciprocity reaches employee compensation only — never self-employment, rent or gains — so `allocated_wage_income` accumulates from `income.type` beside `allocated_income`, and only the wage share is ever exempted
- **`Determine_State_Reciprocity` (40520)** holds the matrix, one row per work state, and removes the exempt wages
- **`Validate_Reciprocity` (8050)** makes the expectation tags real. The OH-KY scenario already shipped asserting reciprocity through three tags that were never mapped and so asserted nothing
- fixes a fencepost in `Allocate_Income_By_State` that counted Jan 1 – Dec 31 as 364 days and allocated 99.7% of a full year's income. Latent because the table had never executed

The relief zeroes the work state's taxable wages but **keeps** the `state_tax_result`: the certificate governs withholding, not liability, so an employee who never filed one still files a non-resident return showing zero wages to recover what was withheld.

## Two exclusions are deliberate

Both are asserted as negatives, because getting them wrong exempts tax that is owed:

- **Minnesota–Wisconsin is terminated** (2010-01-01). Wisconsin Pub 121 (01/26) lists only IL, IN, KY, MI; Minnesota lists only MI, ND. The 2024 joint study required by 2023 Wisconsin Act 147 produced no reinstatement.
- **Arizona is not a reciprocity state**, though widely published as one. A.R.S. § 43-1096 is a withholding exception backed by a credit: the wages stay Arizona-source and Arizona-taxable, a Form 140NR is still required, and the liability is offset on Form 309.

## Contains an engine fix

`fix(el): restart local slots per table row` is a separable commit. Local slot indices are frame-relative, but the counter ran table-wide, so the second row declaring a local emitted `1 local@` against a frame holding one slot and died with `[OutOfBounds] GetFrameValue`. That is #1047 one scope down — #1047 stopped the bleed between tables, this stops it between the rows of one table. It stayed hidden because it needs two rows in one table that each declare a local, and `Build_State_Tax_Result_For_Period` — the first such table anyone executed — was orphaned the whole time it had the defect. Context locals must survive (#965), so the fix is a baseline, not a reset.

## Coverage

- `reciprocity_test.go`: all 30 pairs from **both** directions (58 subtests), 9 negatives, DC's any-nonresident rule, the DC self-exemption guard, and a wage-only case proving rent stays taxable
- four scenarios in `MultiState/` — one a negative (WI→MN), one proving the wage-only limit
- scenario floors 280 → 284, 500 → 504
- `make check` green; `dtrules verify sampleprojects/TaxReturn` clean

Docs: `docs/state-reciprocal-agreements.md`, including the per-pair conditions **not** yet modelled (daily-commuter for KY↔VA, monthly-return-home for MN/ND, 183-day statutory-residence tests, local income taxes, the OH↔KY 20% S-corp carve-out). `docs/non-resident-state-tax.md` gets a status note — much of it described fields and tests that do not exist.

Two unrelated bugs found on the way, filed not fixed: #1168 (`dtrules run --input` discards singleton data) and #1169 (authoring writes into `xml/states/*` skip Excel and break `verify`).

https://claude.ai/code/session_01DMYhZQVsuYAtPFtFwMJoUR


PaulSnow:
Auto-closed when #1170 merged and deleted this PR's base branch. Reopening is refused while that base is gone, so the work continues in #1172 — same commits, rebased onto `main`.

Issue #1209, opened 2026-09-20
feat(edd): field value constraints — allowed_values and max_length/max_words, enforced on every input path


PaulSnow:
## Why

The EDD can describe a field's legal values (`options` on a `multiple_choice` question) but nothing enforces them, and there is no way at all to declare a closed vocabulary or a length limit on a field that is not an interview question. Any write path accepts any string.

**Reproduced on `main` with the flagship sample:**

```bash
dtrules run sampleprojects/SinusitisTherapy --entry Determine_Therapy \
  --input sampleprojects/SinusitisTherapy/testfiles/TestScenarios/AdultStandard/input.xml --save case.xml
sed -i 's#<diagnosis>Acute Sinusitis</diagnosis>#<diagnosis>Banana</diagnosis>#' case.xml
dtrules run sampleprojects/SinusitisTherapy --entry Determine_Therapy --data case.xml
# exit 0 — recommended_drug: Amoxicillin, dose_mg: 500, duration_days: 14
```

A diagnosis of `Banana` yields a confident prescription. The rules fell through to a default column because the value matched nothing; the engine had no declaration to check it against. In `pkg/dtrules/collect/collect.go` the collected value goes straight to `re.Put(attr, value)`; `Options` are copied into the `Request` for display only.

**Second use case — the EDD as a wire protocol.** DTRulesAgents (a project, not yet public) uses a DTRules rule set to supervise AI coding sessions. Sessions report state as EDD-typed records (`stage` ∈ `orient,plan,implement,verify,submit`; `severity` ∈ `high,medium,low`), and decision tables act on them. Two things make that workable: values outside the vocabulary must be *rejected at the boundary* (an AI that invents `stage=almost_done` must get an error, not a silent else-column), and free-text fields must be *bounded* (`summary` ≤ 40 words) so sessions cannot push walls of prose through the coordination channel — the measured cost driver in our runs is tokens re-read, not tokens written. Today we would have to keep a second, parallel schema outside the EDD for both, which defeats "the EDD is the single source of truth."

## Proposal

Optional, per-field, round-tripping through Excel like all other field metadata:

- `allowed_values` — closed vocabulary for `string` (and `integer`) fields. Independent of `collect`; a `multiple_choice` question with no explicit `allowed_values` may default to its `options` values.
- `max_length` and/or `max_words` — for `string` fields.
- Comparison follows EL rules (case-insensitive match; authored spelling preserved).
[...]

Issue #850, opened 2026-06-17
Lazy field collection for interactive execution


PaulSnow:
## Summary

As a decision table executes, when it reads a field that has not been
**collected**, the engine asks for it — inline (CLI) or across a web round-trip —
using per-field question metadata declared in the EDD. Only the fields actually
reached on the executed path are asked. This turns a rule set into a self-driving
interactive interview (a web page, or a CLI prompt sequence) generated from the
rules, with no bespoke UI code — while **batch execution with full data behaves
exactly as it does today.**

## Sub-issues (implementation plan, in dependency order)

- [x] #851 — **EDD: `collect` flag + question metadata** (schema, authoring API,
      Excel round-trip). Foundation; everything depends on it.
- [x] #852 — **Engine: collected/defaulted state + read-point resolver**
      (batch == today). Depends on #851.
- [x] #853 — **Canonical data XML: reader/writer + load modes**
      (authoritative/review). Depends on #851, #852.
- [x] #854 — **CLI interactive resolver** (simplest front-end; validates the
      engine). Depends on #852.
- [x] #855 — **Web resolver + embedded SinusitisTherapy demo** (the capstone).
      Depends on #851, #852, #853.

## Core model: `collected` vs `defaulted`

Every field **instance** is in one of two states:

- **collected** — its value is authoritative. Never asked.
- **defaulted** — it has a value (the EDD `default_value`, or a loaded review
  value), but if the field is flagged `collect` it is **asked**, pre-filled with
  that value.

This is the whole model. It is tracked **per instance** (a specific entity on the
stack — so each `person` in a `household` is independent) and **separately from
the value**, because at runtime an unset field, an explicit null, and the applied
default are indistinguishable (all return the null singleton — see
`pkg/dtrules/entity/entity.go:387`, `null.go`). `collect`/`collected` is metadata,
never inferred from the value.

What is `collected` without asking: constants; values from an authoritative load;
fields **written by an action** (mark-collected on write, `entity.Put`).

---

## Pillar 1 — the blocking read-point resolver

The single field-read chokepoint is `RName` → `state.Find` → `entity.Get`
(`pkg/dtrules/name.go:215`, `pkg/dtrules/interpreter/state.go:528`,
`pkg/dtrules/entity/entity.go:269`). That read is the **one place** with
everything needed to ask: the entity instance, the field, and its question
metadata.

At that point, if the field is `collect` and the instance is `defaulted` (not
collected), the engine calls a resolver:

[...]

PR #1214, opened 2026-09-20
feat(edd): field value constraints — declaration and round trip (#1209)

PaulSnow:
Part of #1209. This is Part A: the declaration and round trip of the new
field metadata. Enforcement on the paths that write a field from outside the
rules (`--input`, `--data`, the collect resolver, the web interview, the API)
is Part B and is **not** in this PR.

## What a field can now declare

```xml
<field name="diagnosis" type="string" default_value="Acute Sinusitis" max_length="40">
    <allowed_value value="Acute Sinusitis"/>
    <allowed_value value="Chronic Sinusitis"/>
</field>
```

- `allowed_values` — a closed vocabulary, on `string` and `integer` fields.
  Independent of `collect`: a field nobody is asked for can still have one.
- `max_length` / `max_words` — bounds on a `string` field, in characters and
  in whitespace-separated words.

Matching follows the rule every name in the system follows: case-insensitive,
authored spelling preserved. `Acute Sinusitis` and `acute sinusitis` are one
value, and what is written back out is what the author wrote.

## The layers it travels through

Following `question_*` (#850) end to end: the entity model
(`entity.FieldConstraints`), the EDD XML loader and writer, the typed
authoring view, `dtrules edd get|put|patch` and both JSON schemas, and the
Excel EDD sheet — columns N–P, `Allowed Values` / `Max Length` / `Max Words`.
`dtrules build` and `dtrules verify` round-trip them byte-identically.

A sheet grows those three columns only where some field on it declares a
constraint. Writing them unconditionally changed the bytes of every workbook
in every project, and an authoring write recompiles a workbook whose bytes
changed — so a one-field edit would have rebuilt the whole project and
attached that diff to it. The importer reads by column position, so an absent
column reads as empty.

## The validate gate

`dtrules validate` reports a field whose own `default` its constraints
reject, and the authoring API refuses to write one. Such a default is
unreachable rather than merely odd: the constraint applies to supplied
values, so the field starts every run holding a value the rules were told it
can never hold, and no input can correct it.

## Tests

The issue's vectors 9, 10 and 11:

- **9** — `dtrules validate` rejects a default outside its own
  `allowed_values`, naming the field, the value and the set; `edd patch`
  refuses the same declaration up front.
- **10** — one `edd patch` lands the constraints in the XML, in the workbook
  and in what `edd get` reports, with `verify` clean afterwards and the
  field's existing `collect` metadata undisturbed.
- **11** — a build from Excel brings them back out of the new columns, and
  XML → workbook → XML is byte-identical.

Plus unit coverage of the declaration rules (which types accept which
constraint, a value listed twice, a limit that is not a number) and of the
patch conventions (omit to keep, `[]` and `"0"` to clear).

`make check` passes.

Spec: new §2.5.1 in `docs/SPEC.md`, and a section in `dtrules docs edd`.


PaulSnow:
Superseded by #1218, which carried this PR's commits (Part A: declaration and round trip) together with Part B (enforcement) and is now merged. Nothing here is lost.

Issue #1234, opened 2026-09-21
Feature (low priority): join <array> by <sep>


PaulSnow:
There is `tokenize`, but no inverse. Joining a string array today needs a loop: `for all arr as x where x != "" { set acc = acc + x + sep }`. That works, but it is verbose and adds a trailing separator. A `join` operator would make that one expression. Low priority: the loop works.

PaulSnow:
**[builder/opus 21:58Z]** Starting on branch `task/1234-join-array`. It is branched from `task/1232-seconds-minutes-time-arith` (PR #1243) because both regenerate the ANTLR parser. The plan is a `join <array> by <strexpr>` string expression and a `join` operator ( array sep -- string ). `join (tokenize s by sep) by sep` must give back s. Test first.

PaulSnow:
**[builder/opus 22:01Z]** Done: PR #1245, branch `task/1234-join-array` @ a7ae9cb1. It is stacked on #1243, so merge that first. Adds `join <array> by <sep>` and a `join` operator. The execution test covers the round trip with tokenize, including empty fields. With the fix reverted, the test fails with a parse error on 'join'. `make check` passes and all sample projects verify. `join` is now a keyword.

PaulSnow:
**[reviewer/fable 22:37Z]** Reviewed PR #1245 @ a7ae9cb1 (stacked on #1243) in throwaway worktrees under /tmp: regen byte-identical, make check green on branch and on the full stack with #1248 (guard sees strJoin), fail-before confirmed, keyword scan for `join` clean, element types/separators/round-trips probed. Verdict: no code defect; one finding: the doc says the separator takes the whole string expression after `by`, but it binds tightly (`join a by sep + x` is `(join a by sep) + x`). Full report as a review comment on the PR.

PaulSnow:
**[builder/opus 22:52Z]** #1245 is rebased onto #1243's fixes (now fe8a976c + 10028f4a). The docs had join precedence backwards: the separator binds as one operand, so `"[" + join w by "|" + "]"` works without parentheses. The docs are fixed and a test pins the precedence.

Issue #1231, opened 2026-09-21
run --save: element order is not stable across identical runs


PaulSnow:
Reproduced on DTRules v1.25.0-20-g6a4bc145 (build/dtrules at 6a4bc145). Running the same table on the same input three times with `--save` gives three different files (different md5). Sorted line by line they are identical, so only the order of top-level elements varies.

That breaks any host that detects "did this run change the state" by comparing saved output: DTRulesAgents has to do a canonical, order-insensitive comparison, which costs 21 ms at 374 KB and 1.5 s at 7 MB. Expected: deterministic output order, e.g. EDD or insertion order.

PaulSnow:
**[builder/opus 21:40Z]** Starting on branch `task/1231-stable-save-order` (off origin/main 6a4bc145). Reproduced: Probe_Ctx on probe-in.xml 3x gives 3 md5s, identical when sorted; also on sampleprojects/SinusitisTherapy. Suspected source: `entity.Factory.GetRefEntities` and `REntity.GetAttributeNames` both range over Go maps, so both the top-level entity order and the field order inside each entity vary. Plan: failing test through `dtrules run --save` first, then return both in EDD declaration order.

PaulSnow:
**[builder/opus 21:42Z]** Done: PR #1236, branch `task/1231-stable-save-order`, commit ae31fdfb. The cause was two map ranges, `Factory.GetRefEntities` (entity order) and `REntity.GetAttributeNames` (field order within each entity, which the sorted comparison hid). Both now return EDD declaration order, using the existing ref-entity id and attribute value index. Three new tests; all fail on main. The PR body lists which ones fail when each half of the fix is removed. On the issue's reproduction, 3 runs now give one md5. `make check` passes (UI vitest skipped, no node_modules).

Issue #1226, opened 2026-09-21
perform: the callee overwrites the caller's local slot 0 (currentFrame never moves)


PaulSnow:
## Summary

A table performed from inside another table overwrites the caller's local slot 0. After `perform`, the caller's context alias (or `local` variable) holds whatever the callee last stored there.

## Cause

Locals are addressed as `currentFrame + index` (`interpreter/state.go`, `GetFrameValue` / `SetFrameValue`). The compiler numbers locals per table, starting at 0. But nothing on the execution path ever moves `currentFrame`:

- `opAllocate` only does `CtrlPush(v)`.
- `opPerformTable` calls `RDecisionTable.Execute`, which evaluates the context directly.
- `PushFrame` / `PopFrame` exist, but have no callers outside two tests (`operators/local_frame_test.go` `TestLocalStoreFetch`, `interpreter/state_test.go`). Those tests call `PushFrame` by hand, so they pass while production runs with `currentFrame` always 0.

So every table's slot 0 resolves to `ctrlStk[0]`. The callee's `allocate` reserves a new slot at index 1, but its `0 local!` writes index 0, the caller's slot. Its `deallocate` then pops the unused slot at index 1.

## Repro

Two tables, run with `dtrules run rules --entry Probe_Caller --data <any state with several state.nodes>`:

Probe_Caller (FIRST): context `for all state.nodes as n where n.id == record.session_id`, one column:
1. `set result.holder = n.id`
2. `perform Probe_Callee`
3. `set result.missing = n.id`

Probe_Callee (ALL): context `for all state.nodes as m where m.id == "no-such-node"`, one action `set result.finding = "callee body ran"`. The table needs a `true` condition to fire, see the separate authoring note.

On v1.25.0-20-g6a4bc145:

```
holder:  e11.4362a        <- n before perform
missing: e11.accept12h    <- n after perform = the last node the callee's loop visited
```

`holder` and `missing` must be equal.

## Why the sample projects pass

`go test ./pkg/dtrules/` is green. No sample uses `as` aliases, since `for all X` uses the entity stack, not a local. SyntaxTests has 9 tables with `local` declarations on both sides of a `perform` (`Run_Test` → `Run_Test_2` → … → `Run_Test_8`, `Run_Test_15` → `Run_Test_16`). In every one, `perform` is the caller's last use of its locals, so the overwritten slot is never read.

## Impact seen

[...]

PR #1235, opened 2026-09-21
fix(el): x matches p tests x against pattern p (#1228)

### LINKED ISSUE #1228: x matches p: operands are inverted, so every documented regex is false
Reproduced on DTRules v1.25.0-20-g6a4bc145 (build/dtrules at 6a4bc145). With `n.id = "e11.4362a"`:

- `n.id matches "^[A-Za-z0-9._-]+$"` → **false** (documented spelling)
- `"^[A-Za-z0-9._-]+$" matches n.id` → **true**

`opRegexMatch` pops the subject as the pattern. Every regex written as documented is silently false, and nothing warns. Regression test: `"abc" matches "^a"` must be true and `"^a" matches "abc"` false.

PaulSnow:
## What was wrong

`s matches p` compiles to `s p regexmatch` (subject pushed first, pattern last). `opRegexMatch` popped the top of the stack as the *subject* and the one beneath as the *pattern*, so the operands were inverted at runtime: `n.id matches "^[A-Za-z0-9._-]+$"` was false and `"^[A-Za-z0-9._-]+$" matches n.id` was true. No warning.

## The fix

In the operator, not the emitter. `opRegexMatch` now pops the pattern first, then the subject: stack effect `( string regex -- boolean )`.

Why the operator: every producer of `regexmatch` postfix in the repo already puts the subject first:
- the Go EL emitter (`VisitBoolMatches`: `<s1> <s2> regexmatch`)
- the legacy compiled postfix stored in `sampleprojects/SyntaxTests/xml/syntaxexample_dt.xml` (`case_ID '22345' regexmatch`, `1 local@ 'This is a string' regexmatch`)
- `docs/el-reference.md` (`taxpayer.filing_status "MF.*" regexmatch`) and `dtrules docs` (`s matches "regex.*"`)

Swapping the emitter's order instead would have left that stored legacy postfix still inverted.

## Callers checked

Every use of `matches` / `regexmatch` in the repo:
- **Go sources**: one implementation (`operators/string.go`) and one emitter site. No other runtime (goruntime, nativeasm, UI) has its own regex match.
- **Sample projects**: only the two SyntaxTests conditions above, both written subject-first (`case_ID matches '22345'`, `Things matches 'This is a string'`), so they now evaluate as written. No test runs them with expected outcomes. None of TaxReturn, CorporateTax, Cribbage, Sinusitis and the rest use `matches` in EL. The `for all vehicles where taxpayer_id matches current taxpayer` in `legacy/taxreturn-split-stalled` is an archived, un-built table.
- **Tests**: `TestRegexMatchOperator` pushed pattern then string, which encoded the inverted contract. It now pushes subject then pattern, with its table unchanged.
- **Docs**: `docs/el-reference.md` named the postfix operator `matches`. It now says `regexmatch` and states which operand is which.

## Tests

- New `pkg/dtrules/matches_exec_test.go` (`TestMatchesSubjectThenPattern`): compiles the EL with `el.Compiler` and executes it through a session, as an action (`set b = …`) and as a condition. It covers `"abc" matches "^a"` → true, `"^a" matches "abc"` → false, and the issue's `id matches "^[A-Za-z0-9._-]+$"` → true (and the reverse → false).
- On `origin/main` both it and the corrected `TestRegexMatchOperator` fail (10 assertions, and 2 subtests). With the fix they pass. `make check` passes.

Fixes #1228


PaulSnow (commented):
## Adversarial review — #1235 `x matches p` (#1228)

**Verdict: no blocking findings.** The fix is the right one and in the right place; I reproduced fail-before/pass-after, mutated nothing that survived, and found no producer of `regexmatch` that pushes pattern-first. Three low/informational items below, none of which should hold the merge.

### What I ran (throwaway worktree `/tmp/rev-1235` at `959f21dc`, base `6a4bc145` = current `origin/main`)

- `make check` on the branch: **passed**. The branch is already on top of `origin/main` (merge-base == main HEAD), so "merged onto main" is the branch itself.
- Fail-before: copied `matches_exec_test.go` and the corrected `TestRegexMatchOperator` onto a clean `origin/main` worktree — 10 assertions fail (`"abc" matches "^a" -> false`, `"^a" matches "abc" -> true`, …) and 2 subtests fail. On the branch both pass, also under `-race -count=3`.
- Every producer of `regexmatch` in the repo is subject-first: the emitter (`VisitBoolMatches`, `postfix_emitter.go:4531`), the two stored rows in `sampleprojects/SyntaxTests/xml/syntaxexample_dt.xml` (4969, 5784), `docs/el-reference.md`, `cmd/dtrules/docs.go:490`. No other runtime has a regex op (grep for `regexmatch`/`"matches"` across `.go/.s/.asm/.c/.rs`: only `operators/string.go` and the EL emitter). No Java source or jar exists in the repo or on this machine, so the legacy engine's pop order is **not verified** from code — the only evidence is that the Java compiler emitted `case_ID '22345' regexmatch` for `case_ID matches '22345'`, i.e. subject-first, which agrees with this fix.
- User-facing flips: I built binaries from main and the branch and ran `dtrules run sampleprojects/SyntaxTests --entry Run_Syntax_Examples --input testfiles/test.xml --trace …`. Traces are identical apart from `arraybind` ordering (a pre-existing map-order nondeterminism, #1231's territory). `Run_Test_15` condition 8 (`case_ID matches '22345'`) **is** evaluated on every run and is `false` before and after with the sample data (case_ID = "22344"). `Syntax_Examples_5` condition 9 is masked `-` in its only column and is never evaluated. So nothing in the repo silently changes result.
- Trial merges (`git merge-tree --write-tree`) against every other open PR branch (#1226–#1234, #1241, #1244): all clean, including the four others that touch `docs/el-reference.md`.

### Findings

1. **Low — an invalid pattern is still silently `false`** (`pkg/dtrules/operators/string.go:318-320`, pre-existing, not introduced here). `regexp.MatchString` error → push `false`, no diagnostic. Scenario: `id matches "^[A-Z"` (unbalanced class) evaluates to `false` forever with no warning — the same "silently false, nothing warns" class this issue is about. Also no compile cache: the pattern is recompiled on every row evaluation. Suggest a follow-up issue: return the error (or at minimum a compile-time check when the pattern is a literal) and cache compiled patterns.

2. **Low — the doc line is looser than the operator.** `regexp.MatchString` is an unanchored search; the new text "True if the left string matches the regular expression on the right" reads as a full match. `"abc" matches "b"` is `true` here. If the postfix format is shared with the legacy Java engine, `String.matches` there is a full match and the same row disagrees. Not verified (no Java source available). Worth one clause: "contains a match for the regular expression; anchor with `^…$` for a full match".

3. **Info — silent flip for anyone outside this repo who worked around the bug.** Any stored postfix or EL written as `"pattern" matches field` to get a true result will now be false with no warning; the PR's "no warning" complaint applies symmetrically to the fix. A cheap advisory in the emitter — left operand a string literal, right operand not — would catch the overwhelmingly-likely inversion at build time. Release-note it either way.

### Test quality

`TestMatchesSubjectThenPattern` compiles the EL and executes it through a session as both action and condition, checks `Execute` errors, and pops the condition result from the data stack — it cannot pass vacuously. It does supply symbols by hand (`SetSymbols({"id": "string", …})`) rather than from an EDD, which is fine for an operator fix since the emitter path is exercised regardless.

### Not verified

- Legacy Java runtime pop order and anchoring semantics (no source available).
- Any postfix stored in projects outside this repository.
- UI vitest suite (`ui/node_modules` absent, `make check` skipped it — same on main).

PR #1236, opened 2026-09-21
run --save: stable EDD declaration order (#1231)

### LINKED ISSUE #1231: run --save: element order is not stable across identical runs

PaulSnow:
## What

`dtrules run --save` now writes entities, and the fields within each, in EDD declaration order. The same rules on the same input save byte-identical files.

## Why

`entity.Factory.GetRefEntities` and `REntity.GetAttributeNames` both ranged over Go maps, so the top-level element order *and* the field order inside each entity varied per run (the issue saw only the first; sorted-line comparison hid the second). A host that detects "did this run change the state" by comparing saved output had to do an order-insensitive comparison.

Order chosen: **EDD declaration order**, not alphabetical and not first-write order.
- It is the order the author wrote the schema in, so the saved file reads like the EDD.
- It is already recorded at no cost: ref-entity ids are handed out as the EDD loader first declares each entity, and each attribute's value index is assigned as `AddAttribute` first sees it. The fix sorts by those, nothing new is stored.
- Insertion (first-write) order would depend on the data and on which rules fired, so two runs over different inputs would lay the same entity out differently.

Because the fix is in the two accessors, every other caller of them (printed results, excel export, EDD tooling) becomes deterministic too. `REntity.GetEntries` and `Factory.GetEntityNames` still range over maps; `--save` does not use them and I left them alone to stay within this issue.

## Tests

- `cmd/dtrules/run_save_order_test.go` drives the real `dtrules run` path (`runRun`, real project load, `--data`, `--save`) on `sampleprojects/SinusitisTherapy`:
  - `TestRunSave_ByteIdenticalAcrossRuns`: 8 runs, byte-identical output.
  - `TestRunSave_EDDDeclarationOrder`: the saved entity order and each entity's field order are subsequences of the EDD's declaration order, parsed from `sinusitis_edd.xml`.
- `pkg/dtrules/entity/declaration_order_test.go`: entities and fields declared out of alphabetical order come back in declaration order, 20 times. (The sinusitis EDD happens to be alphabetical, so this is the test that tells declaration order apart from sorted order.)

Without the fix (both files reverted to main): all three fail. Without only the `GetAttributeNames` fix: all three fail. Without only the `GetRefEntities` fix: the entity test and the byte-identical test fail. The EDD-order CLI test passes there, because its four entities happen to come out in a rotation that matches.

Reproduction from the issue (`Probe_Ctx` on `probe-in.xml`, 3 runs): before, three different md5s; after, one md5 (`b34d41fb…`) three times.

`make check` passes. The UI vitest step was skipped: no `ui/node_modules` in the worktree.

Fixes #1231


PaulSnow:
**[reviewer/fable 22:25Z]** Starting adversarial review of `task/1231-stable-save-order` (ae31fdfb) in a throwaway worktree under /tmp. Plan: check every writer that lists entities/attributes (save, trace, JSON, API), attribute-index uniqueness under the sort, run-time entities, downstream parsers, `make check`, and a trial merge with #1242 (both touch entity/entity.go). Read-only; no push, no merge, no approve.

PaulSnow (commented):
**[reviewer/fable review]** Adversarial review of `task/1231-stable-save-order` @ ae31fdfb, in a throwaway worktree. Not approving or requesting changes; verdict and evidence only.

## Verdict

No blocking findings. The fix is sound, minimal, the tests fail without it, and `make check` passes. The branch sits on the current `origin/main` tip (6a4bc1452), so "merged onto main" is the branch itself. One low-severity leftover (below) is outside `--save` and can stay out of scope.

## What I ran

- `make check` on the branch: passes (35 packages ok; UI vitest skipped, no node_modules).
- Mutation: removed both sorts (entity.go `GetAttributeNames`, factory.go `GetRefEntities`) and re-ran. `TestDeclarationOrder_RefEntitiesAndAttributes`, `TestRunSave_ByteIdenticalAcrossRuns`, `TestRunSave_EDDDeclarationOrder` all fail; restored, all pass. The tests can fail.
- Trial merge with #1242 (`task/1233-run-changed-flag`): clean auto-merge (different hunks in `entity/entity.go`, `doc_embedding.go`); `go test` on `cmd/dtrules`, `pkg/dtrules`, `entity`, `datafile`, `operators` passes on the merged tree.

## Findings

1. **Sort keys are total and stable (verified).** `EntityEntry.Index` is unique per entity: `AddAttribute` assigns `len(e.values)` on first sight and `AddEntry` keeps the old index on replace (entity.go:490-570), so `sort.Slice`'s instability cannot reorder equal keys. Ref-entity ids are assigned under the factory lock in `FindCreateRefEntity`, called from `processEntity` in file order (loader/edd.go:274); the JSON EDD loader iterates a slice (`Entities []JSONEDDEntity`, json.go:74/117), and project files come from `filepath.Glob` (sorted). Declaration order holds for XML and JSON EDDs and multi-file projects.

2. **Every writer that lists an entity now goes through the fixed accessor.** `datafile.Write` recurses through `GetAttributeNames` for nested entities and arrays of entities, and array elements are written in the array's own order (datafile.go:75-130). `trace/writer.go:117`, `trace/report.go:115/194`, `apiserver/debug_entity.go:101`, `debug.go:408`, `server.go:1658`, `interview.go:233`, `run_cmd.go:482` (`renderResult`), `authoring/*`, `excel/exporter.go:803/853` all use `GetAttributeNames`, so trace, printed results and API entity views become deterministic as a side effect.

3. **Low, out of scope: listings still in map order.** `Factory.GetEntityNames` (factory.go:175) feeds `dtrules validate --verbose` (main.go:564) and `session.go:59` (order-insensitive registration, harmless); `REntity.GetEntries` feeds `loader.BuildCollectionIndex` (edd.go:504; only affects the wording of the "ambiguous collection" error, the resolver errors on >1 candidate) and `excel/exporter.go:521`; `Factory.String()`. None touch `--save`. The PR body says so; noting it so nobody assumes `validate --verbose` is stable.

4. **Run-time entities.** `dataEntities` (run_cmd.go:406-424) saves the top-of-stack instance per ref-entity name; instances created by `newentity` appear only when reachable through an attribute or array, in that array's order. No entity type outside the EDD can exist (`CreateEntity` needs the ref entity), so there is no "not in the declaration" case.

5. **Downstream parsers.** `datafile.Read` walks elements by name (order-free), and `--review` uses the same path. DTRulesAgents is external and I did not run it, but the issue says it compares sorted lines, which a fixed order cannot break.

6. **Ordering with #1242.** #1242's `changed` flag is only checkable against bytes once this lands; I ran my #1242 attack table on the 1236+1242 merged tree for that reason. Suggest merging this one first.

## Not verified

UI vitest; the DTRulesAgents consumer; byte-identity of the Excel export.

PR #1237, opened 2026-09-21
Advisory warning: column actions in a table with no conditions never run (#1230)

### LINKED ISSUE #1230: Authoring: a table with columns but no conditions compiles, and its column actions can never run (warn, point at initial_actions)
Reproduced on DTRules v1.25.0-20-g6a4bc145 (build/dtrules at 6a4bc145), with an A/B on the same table. With policy ALL, no condition rows, and one column with an action `set result.table = "nocond-ALL"`, the action never runs. Add a single condition `true` with Y in column 1, and it runs.

That makes an actions-only table (a common shape for a callee that just computes) silently a no-op. Three of our probe tables were silent no-ops until we added `true`. Either fire the column when there are no conditions, or have `validate`/`review` flag the table as an error.

PaulSnow:
## What

A new advisory warning, kind `column actions without conditions`, for column actions that can never run in a table with no conditions. What runs in such a table depends on its policy, exactly as the tree builders decide it:

- **FIRST and ALL:** no tree is built, so no column is ever selected and only the initial actions run. Every action marked in a column is reported:
  ```
  WARN NoCond: has no conditions and policy ALL, so no column is ever selected: action 1 marked in its columns never run — move actions that should always run to initial_actions [column actions without conditions]
  ```
- **BALANCED, which is also what an empty policy loads as:** column 1's actions are the whole tree and run. Only an action marked in later columns and not in column 1 is reported.

The other advisory checks still run on these tables.

## Why

Per the correction on #1230, the engine works as designed. The gap was diagnostic: `table put` accepted such a table and returned no warning. Three DTRulesAgents probe tables were silent no-ops because of it; they were ALL tables.

The check lives in `decisiontable.Analyze`, so `table put/get/patch/warnings`, the MCP table tools, `build` and `review` all report it. `dtrules validate` does not run the advisory pass for any warning kind, so this PR leaves it unchanged.

## After review (415c9b23)

The first version (aa5de736) ignored the policy. It told BALANCED and untyped tables that column 1's actions never run, but they do. It also returned early, which hid the no-op column warning on three TaxReturn skeleton tables (`ky_`, `OK_`, `SC_Military_Retirement_…`). Both are fixed, and those three warnings are back in `review`. No existing warning is removed, and no table in the sample projects triggers the new kind.

## Tests

- `pkg/dtrules/no_conditions_policy_test.go` covers policies `""`, BALANCED, FIRST and ALL:
  - It loads a no-condition table through the real XML loader and executes it in a session. The table has action 1 in column 1, action 2 in column 2 and action 3 in both.
  - It feeds the same XML to the advisory pass the way `build`/`review` do.
  - The warning must name exactly the actions the engine did not run. The engine ran 1 and 3 under `""`/BALANCED, and none under FIRST/ALL.
  - On aa5de736 it fails for `""` and BALANCED.
- `pkg/dtrules/decisiontable/no_conditions_test.go` has unit tests for FIRST/ALL, for BALANCED/none, for a column-1-only table not warning, for the no-op column still being reported, and for one `true` condition stopping the warning.
- `cmd/dtrules/table_no_conditions_test.go` goes through `table put` on a copy of CHIP:
  - policy ALL warns;
  - the same table with no policy does not;
  - the action moved to initial_actions does not.

`make check` passes.

## Docs / spec

- `dtrules docs warnings`: kind 10.
- XML-format topic: a note under `<initial_action_dsl>`.
- `docs/SPEC.md` §2.6.

Fixes #1230


PaulSnow (commented):
**[reviewer/fable 22:50Z]** Adversarial review of `task/1230-no-condition-column-warning` @ aa5de736 (on current origin/main 6a4bc145), in a throwaway worktree.

## Verdict

**Not ready as written.** The warning's central claim — "no column is ever selected and action N marked in its columns never run" — is **false for BALANCED tables**, and BALANCED is what a table loads as when `Type` is empty (`table put` with `policy` omitted writes an empty `Type`). The engine runs column 1's actions in that case. Everything else in the PR holds up: the tests fail with the guard removed, `make check` passes, no textual conflicts with #1235–#1248, and the sample projects don't contain a table that trips the new warning. One secondary inaccuracy in the PR text.

## Findings

### 1. False positive for BALANCED / untyped tables (high)

`pkg/dtrules/decisiontable/analysis.go:91-93` returns the new warning whenever `len(in.Conditions) == 0`, ignoring `in.Policy`. But the engine's tree builders differ:

- `pkg/dtrules/decisiontable/table.go:667-674` (`buildBalanced`): with an empty condition table it installs `NewANodeForColumn(dt, 0)` — **column 1's actions are the whole tree and do run**.
- `table.go:754-758` (`buildUnbalanced`, FIRST and ALL): returns with no tree — column actions do not run.

Default type is BALANCED (`table.go:172`; `SetTypeFromString` leaves it unchanged for `""`/`NONE`).

**Reproduced end-to-end** through the real XML loader and a session (`session.NewRuleSet` → `LoadDecisionTables` → `Execute`), table with `<conditions></conditions>` and one action `set client.eligible = true` marked X in column 1:

```
Type=""         Execute err=<nil> client.eligible=true
Type="BALANCED" Execute err=<nil> client.eligible=true
Type="FIRST"    Execute err=<nil> client.eligible=false
Type="ALL"      Execute err=<nil> client.eligible=false
```

And at the authoring surface: the PR's own `table put` test with `Policy: "BALANCED"` instead of `"ALL"` still receives `column actions without conditions` — telling the author that an action which does run never runs, and to move it to `initial_actions`.

Sample projects contain 2 explicit BALANCED tables (468 tables: 390 FIRST, 76 ALL, 2 BALANCED), none without conditions, so the sweep did not surface this; the JSON put path with `policy` omitted reaches it directly.

Fix options: gate the check on `Policy` being FIRST/ALL (and say so in the warning text, `dtrules docs warnings` kind 10, the `<initial_action_dsl>` note and SPEC §2.6, which all repeat the unconditional claim); or, for BALANCED, warn only about actions marked in columns other than 1 (only column 1 runs there — `NewANodeForColumn(dt, 0)`). Either way the reason text must not say "never run" for a table type in which they do.

### 2. "No existing output changes" is not true (low)

The early return at `analysis.go:91` also returns `nil` when no action is marked, which suppresses `checkNoOpColumns` for every no-condition table. On the sample projects, `dtrules review` on main vs this branch differs in exactly three warnings, all lost on the branch:

```
TaxReturn: ky_military_retirement_exemption col 1 no-op column
TaxReturn: OK_Military_Retirement_Exclusion col 1 no-op column
TaxReturn: SC_Military_Retirement_Exclusion col 1 no-op column
```

These are skeleton tables (Type FIRST, no conditions, one comment-only action with `<columns>-</columns>`). Losing a no-op warning on a skeleton is arguably fine, but it is a behaviour change the PR says does not exist; state it, or keep `checkNoOpColumns` running in the no-condition branch.

## Verified

- `make check`: passes on the branch (35 packages ok; UI vitest skipped, node_modules absent).
- Fail-before: with the `len(in.Conditions)==0` block removed, `TestAnalyze_ColumnActionsWithoutConditions`, `..._Only` and `TestTablePutWarnsColumnActionsWithoutConditions` all FAIL; restored, they pass.
- Loader parity: comment-only condition rows are kept by the loader (`loader/dt.go` → postfix `true always`), so `len(Conditions)==0` in Analyze matches the engine's condition count; no false negative from comment rows.
- Excel import: action columns come from the actions header independently of conditions (`dt_importer.go:1364-1392`), width from both sections (`tableWidth`), so an X survives Excel→XML and the warning fires via `build`.
- Surfaces: `table put/get/patch/warnings` and the MCP table tools all call `analyzeAuthoringTable` → `decisiontable.Analyze`; `build`/`review` call `Analyze` via `buildAnalysisInputs`.
- Sample sweep: `dtrules review` over all 13 sample projects with main and branch binaries; the only `warnings[]` differences are the three above; no `column actions without conditions` fires anywhere.
- Merges: no textual conflicts with any of #1235–#1248 (`git merge-tree` against every open PR branch); merged with #1239 (perform-frame) the el/decisiontable/cmd/pkg tests pass.

## Not verified

- `dtrules validate` behaviour (PR says it runs no advisory pass; not exercised).
- Excel import end-to-end with a real workbook (code-read only).
- Whether BALANCED-with-no-conditions is a shape anyone authors on purpose; the engine runs it, so the diagnostic must at least not misdescribe it.

Issue #1022, opened 2026-08-08
No positional access into entity arrays: `stack.cards[0]` lowers to a bytes index, there is no `first of`, and `for all` has no index


PaulSnow:
Cribbage pegging is the natural next conversion after #984, and it is inherently *positional*: "the last L cards laid, in order". Every #980 primitive is order-independent — subsets, key groups, multiset runs — so pegging has to reach elements by position, and EL cannot.

Probed against the compiler:

```
set peg.top = last of stack.cards      OK   -> stack.cards last getrelationship cve /peg.top xdef
set peg.top = first of stack.cards     FAIL -> no viable alternative at input 'setpeg.top=fir…'
set peg.top = stack.cards[0]           OK   -> stack.cards 0 bytesidx cvi /peg.top xdef
stack.cards[0].rank == 5               FAIL -> token recognition error at: '.r'
for all stack.cards with index i       FAIL -> unexpected tokens after parse: with in
for all stack.cards in reverse         OK   -> dup stack.cards forallr pop              (#975, works)
```

Three separate holes:

1. **`last of` exists, `first of` does not.** `last of` even converts properly to an entity (`last getrelationship cve`). The parity gap is arbitrary.

2. **Indexing an entity array silently lowers to a *bytes* index and an integer conversion.** `stack.cards[0]` compiles to `stack.cards 0 bytesidx cvi`, so `set peg.top = stack.cards[0]` assigns an integer where an entity was meant, with no diagnostic — the documented form is `(string) arrayExpr[iexpr]` / `(long) arrayExpr[iexpr]` (docs/el-reference.md:621), and there is no entity form. Field access on an indexed element does not parse at all.

3. **`for all` has no index.** With one, windowing over a sequence is expressible in EL directly, and a separate windowing primitive may not be needed at all.

## To do

- `first of arrayExpr`, for parity with `last of`.
- Entity-typed element access: `stack.cards[i]` in an entity context and field access on the result (`stack.cards[i].rank`), or an explicit `(entity)` cast form. At minimum the `bytesidx cvi` lowering must be unreachable for an array of entities — a compile error beats a silent integer.
- An index binding in `for all` contexts (`for all <array> with index <name>`), usable in conditions and actions.

## Acceptance

A decision table can reach the last L elements of an entity array and read their fields without host-side help; `stack.cards[0]` on an entity array either yields the entity or fails to compile; `make check` green.

[...]

Issue #869, opened 2026-06-20
EL: boolean aggregations likely share the forall-fold operand bug (#867 follow-up)


PaulSnow:
Follow-up to #867 (which fixed the numeric `sum of` / `number of … where` folds).

The **boolean aggregation** emitters share the same `seed arr { body } forall` shape and therefore almost certainly the same operand-order bug — `forall` is `( body array -- )` and pops the array off the top, so emitting the array before the body block makes it iterate the block's tokens and crash at runtime (`non-Entity entry in array`). The `… where` variants likely also have the `if` operand reversed.

## Suspect emitters (pkg/dtrules/compiler/el/postfix_emitter.go)
- `VisitBoolThereIsInArrayWhere` — `there is <x> in <arr> where <p>` (emits `false arr { bexpr or } forall`)
- `VisitBoolThereIsNoInArrayWhere`
- `VisitBoolAllHave`
- `VisitBoolOneOfHasa`
- `VisitBoolMatchForall`

## Work
1. Add **execution** tests (compile EL, run over real entities, assert the boolean) — the existing tests only assert token presence, which can't see operand order (exactly how #867 shipped broken).
2. Reorder the confirmed-broken emitters to the working `seed { body } arr forall` shape (and `{ block } bexpr if` for predicated ones), matching `VisitForallWhere`.

Found during the #867 review; split out so #867 could close with the numeric-fold fix.

PR #1238, opened 2026-09-21
fix(el): (string) arr[i] reads the array element, not a byte (#1229)

### LINKED ISSUE #1229: (string) arr[i] on a string array fails with RBytesValue: cannot convert to bytes
Reproduced on DTRules v1.25.0-20-g6a4bc145 (build/dtrules at 6a4bc145). `set result.table = (string) node.ancestors[0]`, where `node.ancestors` is a string array holding `["e11"]`:

```
[Conversion Error] RBytesValue: cannot convert to bytes
PostFix: node.ancestors 0  ERROR==> bytesidx  <== cvs cvs /result.table xdef
```

Indexing a string array compiles to `bytesidx` (a byte index) instead of an element index. The same happens for `result.findings[0]`. Expected: the element `"e11"`.

PaulSnow:
## What was wrong

Every cast of an indexed expression (`(string) a[i]`, `(long)`, `(double)`, `(fixed)`, `(boolean)`, `(date)`) goes through `VisitIndxExpr`, and that always emitted `bytesidx`. `bytesidx` only reads bytes: it calls `RBytesValue`. So on any real array it failed:

```
set result.table = (string) node.ancestors[0]
[Conversion Error] RBytesValue: cannot convert to bytes
PostFix: node.ancestors 0  ERROR==> bytesidx  <== cvs cvs /result.table xdef
```

The forms without a cast (`set s = findings[0]`, `set e = ents[0]`) already emitted `getat` when the name is declared `array` (the #1022 and #869 routing). Only the cast forms were wrong.

## The fix

`VisitIndxExpr` now chooses the accessor from the declared type of the indexed name, looking at locals first and then the EDD symbol table (`indexOp`):
- `bytes` → `bytesidx`, which gives the byte, as before.
- anything else → `getat`, which gives the element. The grammar has already placed the name where an array goes, so an undeclared name is treated as an array.

`VisitDateFromArrayAt` had the same hard-coded `bytesidx` and now makes the same choice.

After review, `VisitIntBytesIndex` (the bare `x[i]` form) also goes through `indexOp`. Before, it defaulted an undeclared name to `bytesidx`, while the cast form defaulted to `getat`, so with no symbol table (`CheckAction(el, nil)`, or a build with no EDD) `(long) x[0]` and `x[0]` disagreed. There is now one rule for every form: `bytesidx` for an expression that can only be bytes (hex literal, hash, slice, conversion) or a name declared `bytes`, and `getat` otherwise. `TestIndexAccessorFollowsDeclaredType` pins both forms with no symbols, with an array, with bytes, and for bytes-only expressions. The execution test also runs both forms with nil symbols. On the previous commit, the bare-form cases fail (`x 0 bytesidx`, and `node.counts[1] = 300, want 41`).

The entity-cast assertion now indexes an array whose two elements are different entity types, so an off-by-one index would show.

The EDD symbol table records a field as `array` without its subtype. That is enough: `getat` pushes the element whatever it is, and the cast that follows converts it.

## Tests

- New `pkg/dtrules/array_index_exec_test.go` (`TestCastIndexedArrayElement`). It compiles the way `dtrules build` does: symbols from an `_edd.xml` on disk via `authoring.LoadEDDSymbols`, compiled with `authoring.CheckAction`. It then executes against entities created from the same EDD. It covers:
  - string arrays: the issue's `(string) node.ancestors[0]` → `"e11"`, and index 1
  - number arrays: `(long)` → 300 (not a possible byte), `(double)`, `(string)` → `"300"`
  - boolean arrays and date-string arrays
  - entity arrays: `node.children[1]` gives the same entity, and a `(string)` cast of it works
  - bytes: `(long) node.digest[1]` and `node.digest[3]` still give the byte
- On `origin/main` it fails 7 assertions: string, number, boolean, date and entity casts. With the fix it passes. If `indexOp` is forced to return `getat` for everything, the bytes assertion fails, so the test also guards the bytes case.
- `TestIssue803_DateCastFormsEmitInnerExpr` pinned the bug (`a.dlist 0 bytesidx`). It now expects `getat`.
- `docs/el-reference.md` "Array at index" showed `myarray idx bytesidx cvs`. It now shows `getat` and says when `bytesidx` applies.
- `make check` passes.

## Regenerated sample postfix

`sampleprojects/SyntaxTests/xml/syntaxexample_dt.xml` stored 5 compiled postfix lines that indexed an array with `bytesidx`. They were `(date) clients [2]` twice, `(long) clients [1]`, `( boolean ) flags [ 0 ]` and `( boolean ) errorCodes [ 1 ]`.

I regenerated the file with `dtrules build sampleprojects/SyntaxTests --from-excel`, using a binary built from this branch. The diff is exactly those five lines, `bytesidx` → `getat`. Nothing else in the project changed, and the Excel files are untouched. (A plain `dtrules build` rewrote nothing, because no workbook was newer than its XML.)

**None of those five rows is executed by anything.** They are compiler output brought into line, not behaviour this PR exercised:
- Four are in tables that cannot run: `Syntax_Examples_2`, `Syntax_Examples_4`, `Run_Test_7` and `Run_Test_16`. Each dies on an undefined name first, identically on `main`.
- The fifth, `( boolean ) errorCodes [ 1 ]`, is in `Syntax_Examples_5`. That table does run, but every one of its conditions is `-` in its single column, so the condition is never evaluated. A probe (`elstmterror`) put in that row's stored postfix never fires. The same probe in an evaluated row of `Run_Test_15` does fire.

The array-index behaviour is covered by the tests above, not by SyntaxTests. The broader problem is filed as #1260.

Fixes #1229


PaulSnow (commented):
## Adversarial review — #1238 `(string) arr[i]` reads the element (#1229)

**Verdict: no blocking findings; one low finding and two notes.** The fix does what the issue asks, fail-before/pass-after reproduces, the bytes branch is really guarded (a mutant that always returns `getat` fails the test), and the regenerated SyntaxTests diff is exactly the five lines claimed. But nothing in the repo executes those five regenerated rows — `make check` passing after the regeneration proves nothing about them — and the emitter now has two different defaults for a name of unknown type.

### What I ran (throwaway worktree `/tmp/rev-1238` at `8f0a2dc7`, base `6a4bc145` = current `origin/main`)

- `make check` on the branch: **passed**. Branch is already on top of `origin/main` (merge-base == main HEAD), so the merged result is the branch itself.
- Fail-before: `array_index_exec_test.go` and the updated `issue_803_batch5_test.go` on a clean `origin/main` worktree: 7 assertions fail (`node.counts 0 bytesidx … RBytesValue: cannot convert to bytes`, string/boolean/date/entity casts, and `a.dlist 0 bytesidx`). On the branch they pass, also under `-race -count=3`.
- Mutation: forcing `indexOp` to return `getat` unconditionally fails the test at `(long) node.digest[1]` (`RArrayValue: No Array value exists for this type`, got 300 want 173). The bytes guard is real.
- Regenerated XML: `git diff` is exactly the 5 `bytesidx` → `getat` lines, nothing else.
- Did the regenerated rows execute? **No, nowhere.** `executableSyntaxTables` is `Run_Syntax_Examples, Syntax_Examples_5, Error_Handling_Table`. Of the five tables holding the changed rows, `Syntax_Examples_2`, `Syntax_Examples_4`, `Run_Test_7` and `Run_Test_16` cannot execute at all — run directly with either binary they die on `validatedCitizenship`/`clients`/`expectedChildren` undefined before reaching the changed row, identically on main and on the branch. `Syntax_Examples_5` does execute (it is performed by the entry table) but its only column is `*` for condition 1 and `-` for every other condition, so condition 29 (`errorCodes 1 getat cvb`) is never evaluated: the `--trace` of `Run_Syntax_Examples` shows zero `<condition>` elements for that table. Traces from the main and branch binaries are identical apart from `arraybind` ordering (pre-existing map-order nondeterminism).
- Trial merges (`git merge-tree --write-tree`) against every other open PR branch (#1226–#1234, #1241, #1244): all clean, including #1240/#1243/#1245 which also edit `postfix_emitter.go` and the four others that edit `docs/el-reference.md`.

### Findings

1. **Low — two defaults for a name of unknown type.** `indexOp` (`postfix_emitter.go:5151-5162`) says *unknown → `getat`*; the sibling `VisitIntBytesIndex` (`:1585-1594`) says *unknown → `bytesidx`* via `eexprIsArray`. Both are reachable with no symbols: `CheckAction(el, nil)` is legal ("symbols may be nil", `authoring/el_check.go`), `authoring/table.go` and `execute.go` pass whatever `t.symbols`/`p.symbols` hold, and `cmd/dtrules/build.go:60` only sets symbols when `LoadEDDSymbols(xmlDir)` finds an EDD in that directory. Scenario, no symbols loaded: `set result.count = node.counts[0]` compiles to `bytesidx` and fails on the array, while `set result.count = (long) node.counts[0]` compiles to `getat` and works; conversely `node.digest[1]` works and `(long) node.digest[1]` fails with `RArrayValue: No Array value exists`. Same read, two spellings, opposite outcomes, no diagnostic. Not a regression for arrays (they failed everywhere before) but the PR introduces the second default. One shared helper with one documented default (I'd pick `getat`, arrays being the common case) — or a build-time warning when the base type is unknown — closes it.

2. **Note — the entity-cast assertion cannot detect an off-by-one.** `array_index_exec_test.go:183-187` compares `(string) node.children[1]` with `child1.StringValue()`, which is the entity name `"child"` — identical for `child0`. The uncast assertion just above (`GetID()` of `node.children[1]`) does pin the index, so the test as a whole is sound, but the cast line only proves "an entity was cast", not "the right one". Give `child1` a distinguishing label if `StringValue` can be made to show it, or drop the claim that this line checks the index.

3. **Note — the PR text over-claims on the regenerated rows.** "Stored postfix is what executes, so those lines would still have failed at run time" is true in principle, and "`make check` passes after the regeneration" is true, but per the run above no test or run ever evaluates those five rows, before or after. Worth saying so in the description rather than implying they were exercised. (Semantically they are also odd — `clients` is an array of `client` entities, so `(date) clients[2]` and `(long) clients[1]` will still error at `cvdate`/`cvi` if anyone ever makes those tables runnable; that is the sample's problem, not this PR's.)

### Attack surface checked, no defect found

- `IDENT` lexes at most one dot, so `a.b.c[0]` cannot reach `indexOp`; `alias.field[0]` for a `for all … as alias` local misses `lookupLocal("alias.field")` and falls to `lookupType`'s last-segment lookup — correct.
- `(array) NAME[i]`, `copy of x[i]`, array literals, `tokenize … by …[i]`, and colon-refs all `GetText()` to something the symbol table cannot hold and fall to `getat`, which is right for every one of them except a possessive/colon-ref'd **bytes** field (`:node:digest[i]` / `node's digest[i]` — `GetText()` yields `:node:digest`, which `lookupType` cannot resolve), which would now emit `getat` and fail where `bytesidx` used to work. I could not find any such usage; noting it as the one shape where the new default is wrong.
- No local can ever be declared `bytes` (`declareLocal` is never called with `TypeBytes`), so the `lookupLocal` branch of `indexOp` can only ever choose `getat`; harmless.
- `BaseObject.RBytesValue` errors for every non-`RBytes` type, so no string-typed name ever indexed successfully via `bytesidx` before — no regression for `(long) someString[0]` (errors before and after, with a different message).
- `VisitDateFromArrayAt` uses the same helper; `VisitEntityIndex`, `VisitBoolArrayAt`, `VisitNameArrayAt` were already `getat` and are unchanged.

### Not verified

- The Excel-side of the regeneration (`--from-excel` run itself); I checked only that the resulting XML diff is the claimed five lines.
- UI vitest suite (`ui/node_modules` absent; `make check` skipped it, same on main).
- Behaviour of the debug/speculate API compile paths beyond confirming they load EDD symbols.

PR #1239, opened 2026-09-21
perform: open a local frame per table execution (#1226)

### LINKED ISSUE #1226: perform: the callee overwrites the caller's local slot 0 (currentFrame never moves)

PaulSnow:
## What was wrong

Locals are addressed as `currentFrame + index` (`interpreter/state.go`), and the compiler numbers each table's locals from 0. Nothing on the execution path ever moved `currentFrame`: `PushFrame`/`PopFrame` had no production callers. So a performed table's slot 0 was its caller's slot 0. After `perform`, the caller's context alias or `local` held whatever the callee stored there last. An error part way through a table also left its allocations on the control stack, because the `deallocate` never ran.

## The fix

- `RDecisionTable.Execute` pushes a frame before it evaluates the context or the body, and pops it in a `defer`. That covers `performtable`, `performtableordefault`, `performtableamong`, `performtableamongdefault`, `performcatcherror` (both the main table and the handler), and every entry-point run (`RSession.Execute`, `dtrules run`, interview, the authoring debug wrapper).
- A context runs its own table's body through `executetable` → `ExecuteTable`. That stays inside the frame the context allocated into, which matches the compiler's per-table numbering (context locals first, then row locals from the context's count).
- `performaliased` runs `ExecuteTable` with no context, so it opens its own frame too. EL never emits it.
- `PopFrame` truncates the control stack to the frame base. On the success path that changes nothing; on an error it releases what the table left allocated.
- `PushFrame`/`PopFrame` are added to the `dtrules.State` interface.
- The frame rule is now written in `docs/SPEC.md` §2.4 and `docs/compiler-internals.md`.

## The test

`pkg/dtrules/perform_local_frame_exec_test.go` writes its tables through the authoring SDK, so the postfix comes from the production EL compiler with the production numbering. Every run starts at `RSession.Execute`, and nothing sets a frame by hand. The cases:

- the issue's repro: a caller alias held across a `perform` of a callee that iterates the same collection under another alias
- the same with a `local`
- a caller with 2 locals against a callee with 3, checking that the callee's own locals are its own too
- `perform` nested 3 deep
- a callee that fails inside its context, caught by `performcatcherror`
- a failing entry table, which must leave the control stack and the frame pointer at 0

On origin/main all six fail, for example `result.missing = "n3", want "n1"`, `result.b1 = "n3", want "n1"`, and `control stack holds 2 value(s) after the entry table returned`. With the fix all six pass. `make check` passes.

The DTRulesAgents probe (`dtrules run … --entry Probe_Caller`) printed `holder: e11.4362a` and `missing: e11.accept12h` before the fix. With the fix, both are `e11.4362a`.

`TestLocalStoreFetch` still calls `PushFrame` by hand. It tests the `local@`/`local!` operators only, and its comment now says that and points to the exec test.

Fixes #1226


PaulSnow:
**[reviewer/fable 22:23Z]** Starting adversarial review of this PR. Worktree /tmp/rev-1239 from the PR head (743e555c) on top of origin/main 6a4bc145. Plan: read the diff against the issue and docs, run `make check` on the branch and merged onto origin/main, reproduce fail-before/pass-after by reverting the fix under the new tests, then attack the paths the PR body names. Findings will be posted as a review comment here.

PaulSnow:
**[reviewer/fable 22:29Z]** Milestone: `make check` passes on the branch (branch is one commit on top of current origin/main 6a4bc145, so this is also the merged result). Reproduced fail-before/pass-after myself: with table.go and control.go reverted to origin/main all six new tests fail (`result.missing = "n3", want "n1"`, `result.b1 = "n3"`, `control stack holds 2 value(s)`); restored, all six pass. `go build -tags archive ./...` also passes, so no other State implementer breaks on the interface additions. Now: benchmarks before/after (Cribbage, StateTax), conflicts against the other open PRs, and the paths the PR does not test (conditions/actions declaring locals, performaliased, interview/run/replay).

PaulSnow (commented):
## Review (reviewer/fable) — verdict: **merge**

Adversarial review from a throwaway worktree of `task/1226-perform-frame` @ 743e555c (one commit on top of current `origin/main` 6a4bc145, so the branch result is the merged result). I tried to break it along every path that runs table code and did not find a correctness defect. Nothing below blocks the merge.

### What I ran

- `make check` on the branch: passes. `make check` on a tree with **both** #1239 and #1246 merged onto `origin/main`: passes (35 packages ok, `check passed.`).
- `go build -tags archive ./...` and `go vet -tags archive ./pkg/...`: pass (only the pre-existing `el_parser.go` "unreachable code" vet notes, which neither PR touches). `DTState` is the only `State` implementer in the module, so the two new interface methods break nobody.
- **Fail-before / pass-after, reproduced myself**: with `decisiontable/table.go` and `operators/control.go` reverted to `origin/main`, all six new tests fail with exactly the messages the PR reports (`result.missing = "n3", want "n1"`, `result.b1 = "n3"`, `control stack holds 2 value(s)`, `control stack holds 1 value(s)` under the catch); restored, all six pass.
- **A probe the PR does not have** (written through the same authoring SDK → `LoadFromDirectory` → `RSession.Execute` chain, no hand-set frame, not committed anywhere): a caller whose `perform` sits in an *action column* alongside a `local` declared in that action; a callee with **no context** that declares its `local` in an action column; a caller alias declared in a *condition*'s context and read after two performs. On main: `result.finding = "node"` (the context-less callee's `0 local@` read the caller's alias entity), `result.d1after = "n3", want "n2"`. With the fix: pass. So the frame also covers the `RestoreLocals` numbering (context slots first, row locals from the context count), which the PR's own tests (initial actions only, `Type ALL`) do not exercise. Consider folding a column/condition case into `perform_local_frame_exec_test.go`.
- **Merge conflicts** against every other open PR (#1235, #1236, #1237, #1238, #1240, #1242, #1243, #1245, #1246, #1248): `git merge-tree` is clean for all ten. Shared files are only `docs/SPEC.md` (with #1237, #1240, #1242, #1243, #1245) — different hunks.
- **Performance**, main vs branch, same box, benches run alone, `-count=5/6`:
  - `BenchmarkCribbageScoreHand`: 36.5–39.0 µs/op main, 34.9–44.3 µs/op branch, 593 allocs/op both.
  - `BenchmarkCribbageScorePlay`: 5.8–8.1 µs/op main, 5.6–6.1 µs/op branch, 71 allocs/op both.
  - `-tags archive` StateTax: `AllStates` 1.85–2.48 ms/op main, 1.47–1.53 ms/op branch; `SingleStateTax/NH_High_Income` 808–912 µs main, 721–815 µs branch; `FederalPlusState/Family_2025` 0.95–1.23 ms main, 0.89–0.95 ms branch.
  No regression; the branch is inside or below main's spread everywhere. (There is no TaxReturn-named benchmark in the tree; StateTax + FederalPlusState are the heaviest.)

### Path enumeration (what I checked by reading, with file:line on the branch)

Every table execution funnels through `RDecisionTable.Execute` (`decisiontable/table.go:476`), which now opens the frame before the context and closes it in a `defer`:
- `performtable` `control.go:673`, `performtableordefault` :713/:724, `performtableamong` :804, `performtableamongdefault` :834/:843, `performcatcherror` main :944 and handler :977 — all `.Execute`.
- Entry points: `RSession.Execute`, `dtrules run` (`cmd/dtrules/run_cmd.go:252`), interview (`interview/interview.go:128`), API server (`apiserver/server.go:1621`, `debug_speculate.go:224` via `rsess.Execute`), authoring debug wrapper (`authoring/debug_session.go:297-299` delegates to `inner.Execute`/`inner.ExecuteTable`, so the frame is opened by the inner table exactly as without the wrapper; the fast-forward sentinel returns before any push).
- The only frameless path is the `executetable` operator (`control.go:648`), which is what a context's `{ /T executetable }` wrapper calls — correct, it must share the context's frame. EL emits `executetable` nowhere else (`postfix_emitter.go:3059,3229`), and hand-coded postfix is refused at run time, so a stray `executetable` outside a context is not reachable from a shipped rule set.
- `performaliased` (`control.go:882`) gets its own frame; EL never emits it.
- **Bytecode VM / asm**: `interpreter/vm.go` has no local or frame handling of its own; `local@`/`local!` compile to indexed operator calls that run the same `opLocalFetch`/`opLocalStore` against `DTState.GetFrameValue`, and a table reached from bytecode goes through `OpExec` → `dt.Execute`, so it addresses by the same frame. The VM is only reachable in production from `cmd/dtrules/main.go:603`, which compiles the constant `"1 2 +"`; the rule sessions run the object path.
- **Trace replay** (`trace/trace.go:304,437,480,530`) compiles value bodies and executes them; a table performed inside goes through `Execute`.
- **Error/panic paths**: `PopFrame` is deferred, so it runs on every error return and on panic unwind. There is no `recover()` anywhere in `pkg/dtrules` or `cmd`, so nothing resumes a session after a panic anyway. `PopFrame` truncates `ctrlStk` to the frame base (`interpreter/state.go:546`, unchanged on main), which is what releases the `allocate`s an error left behind — verified by the entry-failure test above.
- Nothing else touches the control stack: the only `CtrlPush` callers are `allocate` (`control.go:575`) and the `>r`/`r>` operators (`stack.go:411/417`), and the EL emitter never emits `>r`; `for`/`forall` use the data/entity stacks only.

### Observations (non-blocking)

1. **New depth limit.** `PushFrame` refuses past `stackLimit = 1000` frames (`state.go:37,533`), so a `perform` recursion deeper than 1000 now fails with `Control Stack Overflow` where before it was bounded only by the Go stack. Reasonable — the data and entity stacks have the same limit — but it is a behaviour change that `docs/SPEC.md` §2.4 does not mention.
2. **Test nit.** `perform_local_frame_exec_test.go:239` calls `st.SetOperatorTable(operators.GetOperatorTable())`. Neither `session.NewSession` nor `dtrules run` does that (it is for the bytecode path); the object path does not need it. Harmless, but it reads as though production requires a step it does not. Not a hand-set frame — the frames all come from `Execute`.
3. The tests are all `Type ALL` tables with initial actions only; see the probe above for the column/condition case.

### Not verified

- The interview/web UI runners end to end in a browser, and the API server under load — only read that they reach `rsess.Execute`.
- The DTRulesAgents `Probe_Caller` reproduction in the PR body (that repo is not on this machine).
- `-tags archive` **tests** (built and vetted only; `make check` does not run them).
- The amd64 `nativeasm` runtime beyond what `make check` runs.

PaulSnow:
**[reviewer/fable 22:35Z]** Review posted as a comment review: https://github.com/DTRules/DTRules/pull/1239#pullrequestreview-5272490052 — verdict **merge**. No correctness defect found on any table-execution path (perform variants, catcherror, entry points, debug wrapper, VM, replay); benchmarks neutral; combined tree with #1246 passes `make check`; clean merge-tree against all other open PRs. Three non-blocking observations (new 1000-frame recursion limit undocumented; test's SetOperatorTable is not a production step; tests cover initial actions only — my extra column/condition probe fails on main and passes here).

PR #1240, opened 2026-09-21
Compile error for a field read after sort … by, with a hint at the name (#1227)

### LINKED ISSUE #1227: sort … by <field>: an entity field compiles but fails at run time (should be a compile error pointing at `the name`)
Found while designing DTRulesAgents tables. Reproduced on DTRules v1.25.0-20-g6a4bc145 (build/dtrules at 6a4bc145).

`sort state.entries in ascending order by entry.key`, written as the docs show, fails at run time:

```
[Undefined] RName: The Name 'key' was not defined by any Entity on the Entity Stack
PostFix: state.entries  ERROR==> entry.key  <== true sortentities
```

`by key` and `by $key` fail the same way. The emitter's sort visitor emits the field as a name lookup (`entry.key`, evaluated before `sortentities`, with no entity on the stack) instead of a literal name (`/key`). The docs' example only works because `name` is a keyword token.

Expected: `sortentities` receives the field name and sorts by it. Regression test: sort a list of entities by a non-keyword string field and by a date field.

PaulSnow:
## What

After `sort … by`, the EL emitter now refuses an identifier whose declared type is known and is not `name`:

```
sort … by entry.key: "entry.key" reads a string value, but `by` takes the name of the field to sort on; write: by the name "key"
```

This covers `by entry.key`, `by key` and `by $key` in ascending and descending sorts, for any non-name type (string, date and so on).

Still accepted, unchanged:
- the keyword `name`
- `the name "…"` and `(name) "…"`
- a field or local declared as type `name`, which holds a sort key
- identifiers whose type the compiler cannot see, for example SyntaxTests' `by $steve`

## Why

Per the correction on #1227, sort works: `by` takes a name expression. The gap was diagnostic. `by entry.key` reads the field's *value*. That read happens before `sortentities` runs, with no entity on the stack, so the table passed `table put` and failed only at run time with "The Name 'key' was not defined by any Entity".

## After review (see the latest commit)

The first version trusted `lookupType`, which falls back to the bare field. The EDD loader fills bare fields last-entity-wins, so with `order.sortkey` typed name and `line.sortkey` typed string, `by sortkey` could be refused.

The check now trusts only three sources: a local, an exact `entity.field`, or a bare name on which every declaring entity agrees. Anything else is treated as unknown and accepted. The error message names where the type came from, for example `(field entry.key)`. `TestSortByBareNameCollision` covers both fill orders and both qualified names, and fails on the first version.

## Tests

- `pkg/dtrules/compiler/el/sort_by_name_test.go`:
  - Five rejected forms (ascending, descending, bare, `$`, a date field). Each must carry `the name "<field>"`.
  - Five accepted forms, with their exact postfix.
- `cmd/dtrules/table_sort_by_test.go`: `table put` against a copy of CHIP, which is the production compile path against a real EDD. It refuses `sort case.clients … by client.gender` with the hint and accepts `by the name "gender"`.
- `pkg/dtrules/sort_by_name_exec_test.go`: executes the hinted forms. `by the name "key"` and `by (name) "when"` sort entities by a string field and by a date field, in both directions. This is the regression the issue asked for.

With `checkSortKey` disabled, `TestSortByFieldReadRejected` and `TestTablePutRejectsSortByFieldRead` fail; the latter reports "put accepted a field read after `by`". The accepted-forms test and the execution test pass either way, because they guard against over-rejection and prove the hint works.

`make check` passes.

## Docs / spec

- The SORT block in `dtrules docs el`: the old `$nameField` example did not say it had to be a name-typed field.
- The sort entry in `docs/el-reference.md`.
- `docs/SPEC.md` §2.3.

Fixes #1227


PaulSnow (commented):
**[reviewer/fable 22:50Z]** Adversarial review of `task/1227-sort-by-name-diagnostic` @ f79bf4da (on current origin/main 6a4bc145), in a throwaway worktree.

## Verdict

**Sound, with one false-reject to fix or at least document.** The diagnostic does what the corrected #1227 asks, the tests fail with `checkSortKey` disabled, `make check` passes, no conflicts with #1235–#1248, and no sample project's output changes. The type lookup it relies on falls back to a bare field name that is last-entity-wins, so a legitimate name-typed sort key can be refused when another entity declares the same field name with a different type.

## Findings

### 1. Bare-name collision refuses a valid name-typed key (medium)

`postfix_emitter.go` `checkSortKey` uses `lookupType(ident)` (`postfix_emitter.go:167-183`), which falls back to the bare last segment when `entity.field` is not in the symbol table. `edd_loader.go:117` fills the bare entry with `symbols[fieldName] = fieldType` for every entity in turn — **last entity wins**. So:

EDD: `order.sortkey` typed `name`, and a later entity `line.sortkey` typed `string` (bare `sortkey` → `string`).

```
sort orders in ascending order by order.sortkey   -> accepted  (direct hit)
sort orders in ascending order by sortkey         -> REFUSED: "sortkey" reads a string value, but `by` takes the name … write: by the name "sortkey"
```

Reproduced on the branch with `NewCompiler()` + `SetSymbols` holding exactly that map. A bare `by sortkey` resolving against the entity on the stack is a legal name-typed read, and the message asserts a type the author did not declare. Field names like `key`, `id`, `name`, `type` recur across entities in real EDDs, so this is not exotic. The symmetric case (bare name resolves to `name` while the intended field is a string) is a false negative — the deliberate "unknown type" hole widened by ambiguity.

Fix options: only trust an exact `entity.field` hit or a local; on a bare-name fallback, treat the type as unknown (accept), or say in the message that the type came from another entity. If the ambiguity is left as is, the PR text "identifiers whose type the compiler cannot see are accepted" should say that bare names resolve to whichever entity was loaded last.

### 2. The deliberate acceptance of unknown types (note, not a defect)

With the EDD loaded, an identifier whose type is unknown is one that is not declared anywhere — it fails at run time regardless of this check — or a `$local` never declared (SyntaxTests' `$steve`). Accepting is the right call; the hole worth noting is only the collision above.

### 3. Attack vectors that turned out moot

- Multi-segment paths (`by case.order.sortkey`) and `as` aliases do not parse in this grammar at all (`IDENT : IDENT_CHAR+ ('.' IDENT_CHAR+)?`; "no viable alternative"), so "fields reached through relationships" cannot reach the check.
- A `$var` local declared as `name` is accepted (`lookupLocal` wins over the EDD type); a local of another type is refused with the hint, which is correct.
- The keyword `name`, `the name "…"`, `(name) "…"`, `nameOf`, `using` and the other `nexpr` alternatives are untouched (only `nameTyped` and `nameLiteral` are inspected).

## Verified

- `make check` passes on the branch (35 packages ok; UI vitest skipped).
- Fail-before: with both `e.checkSortKey(ctx.Nexpr())` calls commented out, `TestSortByFieldReadRejected` and `TestTablePutRejectsSortByFieldRead` FAIL; `TestSortByNameAccepted` and the execution test pass either way, as the PR states.
- Surfaces: `table put` exercised by the PR's test; `table patch` and the MCP table tools go through the same authoring compile (`cmd/dtrules/mcp_tools.go`, `table_cmd.go`); Excel import compiles actions with the same `elCompiler.CompileAction` (`excel/dt_importer.go:1665`) and records the error as a drop. Only `table put` was executed; the rest is code-read.
- Sample sweep: `dtrules review` over all 13 sample projects, main vs branch binaries — no difference in `warnings[]` or `errors[]`. Note the samples contain only `by $steve` sorts (SyntaxTests), so the sweep proves no regression, not that the check was exercised on a real project.
- Merges: no textual conflicts with any of #1235–#1248; merged with #1246, #1238, #1235 and #1239 individually, the el/decisiontable/cmd/pkg tests pass.

## Not verified

- MCP and `table patch` were not executed with a bad `by`.
- Excel import end-to-end with a real workbook containing a bad `by`.
- What the runtime does with a `$name` literal in postfix (`by $sort_field` emits `$sort_field` raw); the accepted-forms test only checks the postfix string.

PR #1242, opened 2026-09-21
Report whether a run changed the state (#1233)

### LINKED ISSUE #1233: Feature: report whether a run changed the state (dirty flag)
A host that runs rules on a timer (DTRulesAgents runs a tick every ~5 s) needs to know whether a run changed anything, so it can skip journaling and tracing no-op runs. Without that, a 5 s tick writes about 17,000 no-op lines a day. Comparing the saved output is expensive and needs order-insensitivity (see the --save ordering issue).

Asked for: a flag in the run result or `--save` output, e.g. `changed=true|false`, set when any entity attribute or collection was written with a different value.

PaulSnow:
## What

The engine now records whether the running rules changed entity data. A host running rules on a timer can skip journaling a no-op run without diffing any output.

- **Go API:** `dtrules.ChangeTracker` (`Changed`, `ResetChanged`, `MarkChanged`), implemented by `interpreter.DTState`. Call `ResetChanged()` after the data load, then `Execute`, then `Changed()`.
- **`dtrules run --save`:** the root element carries the flag, `<dtrules-data changed="true|false">`. `datafile.Read` ignores root attributes, and a test checks that saved files still replay. `datafile.WriteRun` writes the attribute for embedded hosts; `datafile.Write` is unchanged.
- The CLI resets the flag after loading data, so input loaded through a mapping is not counted as the run's change.

**Depends on #1236.** Branch `task/1231-stable-save-order` is merged in because the byte oracle below needs a stable save order. Merge #1236 first; this diff then shrinks to #1233 alone.

## Design

The engine tracks writes; it does not diff state.
- **Attributes.** Every rule write goes through `DTState.Def`, as do collector answers (checked in `Find`). A write counts when the new stored value would save differently from the old one: a different type, different text, a different entity (by identity), or arrays that differ pairwise. This is `dtrules.SameValue`. It is deliberately stricter than `Equals`, which lets doubles differ by 1e-9.
- **Collections.** The in-place array operators count only an effective change. That covers add, addat, remove, removeat, clear, copyelements, addarray and add_no_dups. The sorts, `randomize`, argmax/argmin and the combinatorics destination arrays count only when their elements differ afterwards.
- **Array literals.** `newarray` marks its array fresh until an attribute or another array holds it, so `set x = ["a","b"]` is judged by the `Def` that stores it.
- **Entity stack.** The stack decides which instance `--save` writes. `ResetChanged` records it, and `Changed()` also reports true if it differs, so an unbalanced `entitypush`/`entitypop` counts. A balanced push/pop does not.

**What is guaranteed:** if the running rules changed what `--save` writes, the flag reads changed. The reverse does not hold. The flag records writes, not a diff, so 3 → 7 → 3 reads changed, and storing an entity created during the run always counts.

**Not counted:** writes that bypass the rules. These are `entity.Put` from a Go host, the mapping and datafile loaders (input; the CLI resets after them), the authoring surfaces (`authoring/execute.go`, the debug session), the API server, and trace replay. `populateErrorEntity` uses `Put`, but only on a new error entity it pushes, so the stack comparison catches it. All of this is in `docs/SPEC.md` §2.4/§2.11, `dtrules docs embedding` and `run --help`. `docs/api-guide.md` is the legacy Java guide and was left alone.

### Review findings (Fable), addressed in 59c3485e
1. Stack shape was untracked: pushing a new instance or popping gave changed="false". Fixed by the stack comparison. Tests: `Push_New_State`, `Pop_State`, `Push_Pop_Balanced` (CLI) and `TestChangeTracker_EntityStack`.
2. `Equals` is looser than the saved text. Fixed with `SameValue`. Tests: `Nudge_Ratio` (3 → 3.00000000001) and `TestSameValue_IsStricterThanEquals`. The name-case variant does not reproduce through the engine. Names intern case-insensitively and the first spelling wins, so `/ABC` over `abc` saves as `abc`: nothing changed, and the flag says so. The `Recase_Tag` case pins this, with the byte oracle agreeing. `SameValue` would catch a case change if one ever reached the saved text.
3. The docs now state the guarantee correctly and name the bypass paths. The earlier "errs only toward changed" was wrong.

## Tests

- `cmd/dtrules/run_changed_test.go` goes through the production path: project load, mapping, `--data`/`--input`, `--save`. Each case has an expected flag, and a **byte oracle** checks it: the saved data (attribute stripped) must differ from a no-op run's exactly when the flag says changed. Cases:
  - unchanged: no-op, same value, `x = x`, equal literal, remove absent, balanced push/pop, same double, same name, recased name;
  - changed: one field, reordered literal, add, remove, remove-at, clear, push new instance, pop, nudged double.
  - Also: a saved file replays, and mapped `--input` alone reads unchanged.
- `pkg/dtrules/run_changed_test.go` covers the Go API: execute cases, sticky-until-reset, sort already sorted vs sort that reorders, scratch array, removing absent elements, collector answer, entity stack, and `SameValue` vs `Equals`.

Each part of the fix was removed on its own; something fails every time:

| removed | fails |
|---|---|
| Def comparison marks nothing | Change_Count, Reorder_List, Execute/Change_Count, Sticky |
| Def marks every write | Same_Count, Self_Count, Same_List, Execute/Same_*, Sticky, CollectedAnswer |
| `newarray` not fresh | Same_List (CLI+API), scratch array |
| array ops never mark | Add_Item, Remove_Item, Remove_At, Clear_List, Execute/Add_Item, clear |
| `WatchArray` no-op | sort reorders |
| collector not compared | CollectedAnswer |
| CLI does not reset after load | MappedInputIsNotTheRunsChange |
| stack not compared | Push_New_State, Pop_State, EntityStack push/pop |
| `SameValue` falls back to `Equals` | Nudge_Ratio, SameValue_IsStricterThanEquals |

`make check` passes on this branch, which includes #1236. The UI vitest step was skipped: no `ui/node_modules`. The CLI and API tests passed with `-count=3`.

Fixes #1233


PaulSnow:
**[reviewer/fable 22:25Z]** Starting adversarial review of `task/1233-run-changed-flag` (4870e1fd) in a throwaway worktree under /tmp. Plan: enumerate every write path that bypasses `DTState.Def` (xdef, setat/setelement, tables, entity creation, collector/SDK writes, mapping) and build rules that change state through each, hunting false `changed="false"`; check the fresh-array heuristic; measure the Def compare on the heaviest benchmark; `make check`; trial merges with #1236 and #1240. Read-only; no push, no merge, no approve.

PaulSnow (commented):
**[reviewer/fable review]** Adversarial review of `task/1233-run-changed-flag` @ 4870e1fd, in a throwaway worktree. Not approving or requesting changes; verdict and evidence only.

## Verdict

The tracked paths are correct: every rule write I could construct through `Def`, the array operators, nested entities, entity-typed attributes, arrays of entities and the fresh-array literal path reports the right answer. But the statement in §2.4, the embedding guide and the PR body that the flag **"errs only toward changed"** is false. I reproduced three runs whose `--save` output differs while `changed="false"`, which is the dangerous direction for a host that skips work on it. None is a compiled-EL everyday case; all should be either fixed or stated as limits before this merges. Details and evidence below. `make check` passes; the tests can fail; the cost of the compare is unmeasurable.

## What I ran

- `make check` on the branch (on the `origin/main` tip 6a4bc1452): passes; UI vitest skipped.
- Mutations of `DTState.Def`: compare removed (never marks) -> `Change_Count`, `Reorder_List` fail; compare replaced by always-mark -> `Same_Count`, `Self_Count`, `Same_List` (CLI) and `TestChangeTracker_Execute/Same_*`, `_StickyUntilReset`, `_CollectedAnswer` fail. Restored, all pass.
- An attack table of 19 tables through `RSession.Execute` with a byte oracle: canonical save (same entities the CLI's `dataEntities` picks) before and after each table, compared with `Changed()`. Run on the **1236+1242 trial-merged tree**, because on this branch alone the save order is still random and byte diffs are meaningless. 12 controls consistent; 3 false-unchanged; 3 false-changed of the documented kind.
- Trial merges: with #1236 clean (tests pass); with #1240 (`task/1227-sort-by-name-diagnostic`) clean auto-merge on `docs/SPEC.md`, tests pass on `cmd/dtrules`, `pkg/dtrules`, `operators`, `compiler/el`, `interpreter`. #1240 does not touch `operators/array.go`, so no sort-operator conflict.

## Findings, by severity

**1. Medium — the entity stack changes what `--save` writes, and nothing tracks it.** `dataEntities` (run_cmd.go:406-424) saves whichever instance `FindEntity` finds on top of the stack; the only tracked writes are `Def` and the array operators.
- `/state newentity entitypush` (new instance left on the stack): before, the saved `<state>` had `count`, `items`, `ratio`, `sub`, `nodes`; after, it has only the defaults `<count>3</count><ratio>3</ratio>` — `changed="false"`.
- `entitypop` (the loaded instance popped): after, the file is an empty `<dtrules-data></dtrules-data>` — `changed="false"`.
A host that journals on the flag skips exactly the run whose saved state vanished. Reachable only through hand postfix or an unbalanced table today (EL loops push and pop in pairs), so I rate it medium, not high. Cheapest fix in the CLI: compare the ids of the instances `dataEntities` returns before and after `Execute` and mark on a difference; or state in §2.4 that the flag tracks attribute and array writes, not the stack.

**2. Low — `Equals` is looser than the saved text.** `RDouble.Equals` is `|a-b| < 1e-9` (double.go:25,133) while `StringValue` prints full precision (double.go:154). `3.00000000001 /state.ratio xdef` over `ratio=3`: saved `<ratio>3</ratio>` becomes `<ratio>3.00000000001</ratio>`, `changed="false"` (reproduced). Same class, not reproduced (my name field defaulted to null): `RName.Equals` is `EqualFold` (name.go:268), so a case-only rewrite of a name attribute saves different text and reads unchanged. Either compare `StringValue` for doubles and names in `SameValue`, or list this under the limits.

**3. Low — "every rule write goes through `Def`" is true; the spec should say host writes are not counted.** Writes that bypass `Def`, all host-side: `collect.go:143` (compared only when it happens inside `Find`; a host calling `MaybeCollect` or `Put` itself is untracked), `authoring/execute.go:252`, `authoring/debug_session.go:166-176/404/709`, `apiserver/server.go:1807/1850`, `trace/trace.go:287/315/383` (replay), `populateErrorEntity` (control.go:996, onto a scratch error entity that is popped). No rule path bypasses `Def` — I checked `xdef` (stack.go:549 -> `Def`), the VM `def` (vm.go:408), the asm helper (asm_helpers.go:184), `getrelationship`/`hasrelationship` (reads only), `merge`/`intersection`/`deepcopy`/`split` (new arrays), and there is no set-at-index operator or table type. One sentence in the embedding guide would close this.

**4. Info — fresh-array heuristic.** Create, store, remove in one run behaves: `newarray` -> fresh; `Put`, `Add`, `AddAt` adopt (array.go:449-467, entity.go:466-469); once adopted, `addto`/`remove`/`cleararray` mark. `newarray dup state.items copyelements /state.items xdef` (copy into a fresh array, store it back) correctly reads unchanged. One asymmetry: `RArray.Add` adopts the element even when the *holder* is itself fresh, so `newarray dup newarray addto dup 0 getat "x" addto pop` marks changed although nothing is held — false-changed only, and compiled literals fill the inner array before adding it, so it does not arise from EL.

**5. Info — no measurable cost.** Microbenchmark on the merged tree (5 runs each, same allocations): `set state.count = 3` with the compare active 169-189 ns/op vs skipped (flag pre-set) 179-206 ns/op; `set state.items = <equal list>` 348-383 vs 356-377 ns/op. `BenchmarkCribbageScoreHand`, 6 runs: main 41-70 µs/op, branch 43-76 µs/op — noise. Note the PR's cribbage figure cannot see the compare at all: the flag is sticky, so after the first changing write in a reused session every later `Def` skips it; the microbench isolates it. Under `--trace` the CLI attaches a recorder collector (run_cmd.go ~L95), so `Find` now does two extra `Get`s per field read until the first change — bounded, not measured.

**6. Pre-existing, found on the way — `sortentities` nil-pointer panic.** `state.nodes /val true sortentities` panics at operators/array.go:401 (`v1.Compare(v2)` with `v1 == nil`): `e.Get(name)` returns `(nil, nil)` for the non-executable name form. Same code on main; this PR only adds the `defer WatchArray` at :374. Worth its own issue; #1240 is in the sort-by area.

**7. Info — confirmed as claimed.** `datafile.Read` matches the root by `Name.Local` only (datafile.go:184), so `changed=` is ignored; `datafile.Write` is unchanged; the CLI resets after `--data`/`--input`/`--review` load and before `Execute` (run_cmd.go:252-256); `WatchArray` copies the element slice before the in-place sort and the `defer f()()` form evaluates the snapshot at the right time.

## Attack table (merged tree)

| table | bytes differ | Changed() | |
|---|---|---|---|
| set count = 3 / = 7 | no / yes | false / true | ok |
| nested `state.sub` val same / different | no / yes | false / true | ok |
| `state.sub` re-stored (same object) | no | false | ok |
| new `node` added to `state.nodes` | yes | true | ok |
| `sortentities` reorders / already sorted | yes / no | true / false | ok |
| `removeat` out of range | no | false | ok |
| copy into fresh array, store back | no | false | ok |
| bool null -> false | yes | true | ok |
| in-place `addto` then self-assign | yes | true | ok |
| **`newentity entitypush` left on stack** | **yes** | **false** | **false unchanged** |
| **`entitypop`** | **yes** | **false** | **false unchanged** |
| **double within 1e-9** | **yes** | **false** | **false unchanged** |
| clear then refill same | no | true | documented |
| Def on a pushed-then-popped scratch | no | true | documented |
| fresh inner adopted by fresh outer | no | true | finding 4 |

## Not verified

UI vitest; the DTRulesAgents host; the interactive CLI asker; the `--trace` recorder cost; JSON and API-server hosts; the `RName` case variant of finding 2.

PR #1243, opened 2026-09-21
EL: time differences and offsets below a day — seconds/minutes from, d + N seconds (#1232)

### LINKED ISSUE #1232: Feature: time differences in seconds (seconds/minutes from d1 to d2, d plus N seconds)
Date arithmetic today is `adddays`, `daysbetween`, `monthsbetween` and `yearsbetween`. `get hour/minute/second of d` read components, and `is after` compares at second resolution. There is no difference or offset below a day: `days from` two times on the same day is 0.

Use case (DTRulesAgents): a scheduler's rules need "idle for more than 120 s" or "no progress for 300 s", computed from record timestamps. Today the host has to stamp an integer epoch field so the tables can compare integers.

Asked for: `seconds from d1 to d2` (and minutes), `d plus N seconds`.

PaulSnow:
## What

EL gains time arithmetic below a day:

| EL | postfix | result |
|---|---|---|
| `seconds from d1 to d2` | `d1 d2 secondsbetween` | integer, whole seconds |
| `minutes from d1 to d2` | `d1 d2 minutesbetween` | integer, whole minutes |
| `d + N seconds`, `d - N minutes` | `d N addseconds`, `d N negate addminutes` | date |
| `add N seconds to d`, `subtract N minutes from d` | same ops | date expression, and a mutating statement like the day forms |

## Semantics and time zones

These operations measure elapsed time between instants. A date value is an instant that carries a zone.

- A difference is `d2 - d1`: negative when `d2` is earlier, truncated toward zero (119 s is 1 minute, -119 s is -1).
- The zone a date carries does not affect the result. `2026-04-17T21:05:30Z` and `2026-04-17T16:06:30-05:00` are 60 seconds apart.
- Across a daylight-saving change, the result counts the time that actually passed. In Chicago, 00:00 to 03:00 is 7200 s on the spring-forward day and 14400 s on the fall-back day.
- `d + 3600 seconds` from 01:30 CST is 03:30 CDT.
- An offset keeps the zone of the date it moved.
- `days from` is **not** a calendar distance: `daysbetween` counts whole 24-hour periods. The docs and SPEC now say so, with the DST consequence (Chicago Mar 8 to Mar 10 is 1). Behaviour is unchanged; whether it should count calendar days is #1265.
- The operators compute on Unix seconds rather than `time.Duration`, so spans beyond about 292 years stay exact.
- An offset whose sum overflows, or lands outside years 1-9999, is an error, not a wrapped date. So is a difference too large for an int64 of seconds. (Fixed after review, in d9ff9087. At d7fdd2fc only the minute multiplication was checked.)

## Grammar changes and clashes

- New tokens: `MINUTES` (`'minute' 's'?`) and `SECONDS` (`'second' 's'?`).
- 14 labelled alternatives that mirror the day forms (datestatement, dexpr, iexpr).
- The parser was regenerated with `java -jar antlr-4.13.1-complete.jar -Dlanguage=Go -visitor -package el EL.g4`. Regenerating from the unchanged grammar gives byte-identical output, so the diff in the generated files comes only from this change.
- `second(s)` and `minute(s)` are now keywords. A bare field named exactly `seconds` would have to be written with its entity; `job.seconds` lexes as one identifier.
- No sample project or test fixture uses either word as an identifier. `dtrules verify` passes on every sample project with the new lexer (`KidAid_Application` has no xml/excel dir and is not a project).

## Decided here

- **`d plus N seconds` in the issue is spelled `d + N seconds`.** EL has no `plus` word. The docs said `d plus N days`, but that form fails to parse on main. I corrected the date sections of `dtrules docs el`, `dtrules docs operators` and el-reference to `d + N …` rather than making `plus`/`minus` keywords. That would change every `+` alternative in the language and reserve two more words, which is a separate decision.
- **Seconds and minutes only**, as the issue asks. Hours would reserve `hour(s)`.
- **The docs now say `current date` is today at midnight, not now.** For a scheduler measuring against the clock, the current instant is `current date in zone "UTC"`. A test pins that this works with `seconds from`.

## After review (d9ff9087)

- Overflow: `addseconds`/`addminutes` check the sum and the year range. `secondsbetween`/`minutesbetween` check the difference. New tests: `TestAddSecondsSumOverflow` and `TestSecondsBetweenOverflow`. Both fail on d7fdd2fc (`addseconds MaxInt64` returned year 292277026652 with no error).
- Docs: SPEC §2.4, `docs el`, `docs operators`, el-reference and the `opCurrentDateInZone` comment now agree:
  - `current date` is midnight UTC (`today`).
  - `current date in zone` is the current instant.
  - The el-reference `current time` form, which never parsed, is removed.
  - An explicit `now` is #1266.
- The docs state that the keywords are reserved in any case and that leap seconds are not counted.

## Tests

- `pkg/dtrules/time_seconds_exec_test.go`: EL compiled and executed. It covers:
  - same day, across midnight, month end, year end and a leap day
  - pure dates, negative spans, truncation toward zero, sub-second truncation
  - mixed UTC offsets, 400 years
  - offsets across year and month ends, negative counts, a count from a field
  - zone preservation (compared as text)
  - `add`/`subtract` statement mutation, a round trip, the issue's "idle > 120 s" across midnight, and `current date in zone "UTC"`
- `pkg/dtrules/operators/datetime_seconds_test.go`: spring-forward and fall-back in America/Chicago, a zone against UTC, an 800-year span, and minute overflow.
- The grammar sweep has a row for every new label, and `TestAllOperatorsRegistered` lists the four operators.

With the grammar, emitter and operators reverted to main, every new test fails: the exec tests, the sweep rows and the operator tests. A `time.Duration`-based `secondsBetween` fails the 800-year and 400-year cases.

`make check` passes.

## Not in this PR

`TestGrammarSweep_CoverageGuard` currently extracts **0** labels from EL.g4, because its rule-start regex expects `name :` on one line. It passes vacuously for every label, not just these. That is out of scope here; filed as #1244.

Fixes #1232


PaulSnow (commented):
**[reviewer/fable]** Adversarial review of `task/1232-seconds-minutes-time-arith` @ d7fdd2fc, in throwaway worktrees under /tmp (nothing pushed or changed).

## Verdict

No blocking code defect. `make check` passes on the branch, on `main`+#1243, and on `main`+#1243+#1245+#1248 (all clean merges; #1248 is an ort auto-merge of `grammar_overrides.tsv` with no conflict). The regenerated parser is byte-identical to regenerating from the branch's `EL.g4` with the 4.13.1 jar (all 10 generated files). Every new test fails on `main` with only the tests copied in (7 exec tests: parse errors on `seconds`/`minutes`; 4 operator tests: nil-op panic; 14 sweep rows: COMPILE_ERR). With #1248 merged, the coverage guard enumerates 623 labels and fails when I delete the `intSecondsBetween` row, so the new labels are really guarded there — on this branch alone the guard is still vacuous, as the PR says.

Three findings, none in the arithmetic itself: one silent overflow the PR body says cannot happen, and two places where the new spec/doc text contradicts what the code does.

## Findings

### 1. `addseconds`/`addminutes` still return a wrapped date on overflow (Low; contradicts the PR body)

`pkg/dtrules/operators/datetime.go:474` — `addSecondsTo` computes `t.Unix()+secs` with no check. `opAddMinutes` (line 543) guards `n*60`, but not the addition that follows.

Reproduced at the operator level:
```
addseconds MaxInt64          -> 292277026652-12-05 15:30:07 UTC, err=nil, result.Before(start)=true
addminutes MaxInt64/60       -> 292277026652-12-05 15:30:00 UTC, err=nil, result.Before(start)=true
```
`MaxInt64/60` passes the minute guard, `n*60` does not overflow, `t.Unix()+n*60` does. The PR body says "A minute count too large to convert to seconds is an error, not a wrapped date" — it is an error only for the multiplication. Scenario: `d + n seconds` with `n` from an integer field holding garbage; the rule gets a date, not an error. Fix is a bounds check on the sum in `addSecondsTo` (or a shared helper both ops call). Non-overflowing large spans round-trip exactly (2^60 s: `secondsbetween` gives 1152921504606846976 back).

### 2. New SPEC text says `days from` is calendar arithmetic; the code says elapsed/24h (Medium — spec contradicted by code)

`docs/SPEC.md:181` (added here): "Calendar arithmetic (`d + N days|months|years`, `days|months|years from`) works on the calendar". `cmd/dtrules/docs.go:635`: "Use days/months/years from for calendar distances."

`opDaysBetween` (`datetime.go:460`) is `int(t2.Sub(t1).Hours() / 24)`. Measured in America/Chicago:
```
Mar 8 00:00 -> Mar 9 00:00  (1 calendar day, 23h): seconds=82800  seconds/86400=0  daysbetween=0
Mar 8 00:00 -> Mar 10 00:00 (2 calendar days, 47h): seconds=169200 seconds/86400=1  daysbetween=1
Nov 1 00:00 -> Nov 2 00:00  (1 calendar day, 25h): seconds=90000  seconds/86400=1  daysbetween=1
```
Same in Europe/London (82800 s -> 0 days) and Australia/Sydney and Lord_Howe (30-minute shift). So `seconds from d1 to d2 / 86400` agrees with `days from d1 to d2` everywhere — good, they are the same quantity — but neither is calendar. The paragraph as written tells a rule author that `days from` counts calendar days, and it does not (`adddays` is calendar via `AddDate`; `daysbetween` is not — a pre-existing asymmetry this PR now writes into the spec as if it were consistent). Either narrow the sentence to `months|years from` and record `daysbetween` in DIFFERENCES.md, or fix `daysbetween`; that is a decision for Paul, not this PR, but the spec sentence should not ship false.

### 3. `current date` doc change is half-applied and codifies a wart (Low)

The PR adds "`current date` is today at midnight, not the current instant; for now write `current date in zone "UTC"`". Verified: `current date` -> `today` -> `2026-09-21` (midnight UTC); `current date in zone "America/Chicago"` -> `2026-09-21T17:32:31-05:00` (the instant). But on the same branch:
- `docs/el-reference.md:484-487` still says `current date` "Returns the current date/time. Postfix: `currentdate`" (it emits `today`).
- `cmd/dtrules/docs.go:2325` still says `current date in zone "America/Chicago"` is "today in Chicago" — the opposite of what line 639 now says about the same form.
- `datetime.go:110` comment on `opCurrentDateInZone`: "Same instant as `current date`" — false.

The recommended "now" idiom rests on `current date` and `current date in zone` meaning different things (midnight vs instant). That is the asymmetry a scheduler author will trip on; `now` is already a registered operator with no EL surface. Worth a decision rather than documentation.

## Checked and clean

- Keyword clash: scanned every EDD/DT XML, xlsx `sharedStrings`, test fixtures, website, legacy and `pkg/` testdata for `second(s)`/`minute(s)` as identifiers (case-insensitive, word-bounded, not dotted): every hit is a comment or prose. `dtrules verify` passes on all 11 sample projects with the new binary, and the built XML embeds compiled postfix, so that really recompiled every table.
- Lexer: `secondary`, `minutes_idle`, `seconds2`, `clock.d1` still lex as IDENT; bare `second`/`minute`/`seconds`/`Seconds` fail to parse in any case (grammar is `caseInsensitive`, so the new reserved words are reserved in every case — worth one line in the docs). `get secondof d` unaffected.
- Semantics: DST in Chicago, London, Sydney, Lord_Howe, Kolkata all give elapsed time; pre-1970 with nanos truncates toward zero across the epoch; 1900-1969 span exact; sub-second truncation correct in both signs; zone kept for named and fixed-offset zones; `add 5 seconds to clock.d1` (qualified) works; chained `d + 5 seconds - 3 minutes` works; `d + n * 2 seconds` works.
- Leap second: Go ignores them — `2016-12-31T23:59:59Z` to `2017-01-01T00:00:00Z` is 1, and the engine's date parser rejects `23:59:60Z` outright. Consistent with every other date op here; a one-line doc note would do.
- `seconds from d1 to d2 + 5 days` binds as `seconds from d1 to (d2 + 5 days)`, same as `days from`. Fractional counts truncate silently (`d + 1.5 seconds` -> +1 s, `+ 2.9 minutes` -> +2 min), same as the day forms.
- gofmt clean on all hand-written files.

## Not verified

UI vitest (no node_modules); behaviour with tzdata absent; nothing beyond what `make check` covers for the amd64 assembly executor.

PR #1245, opened 2026-09-21
EL: join <array> by <sep>, the inverse of tokenize (#1234)

### LINKED ISSUE #1234: Feature (low priority): join <array> by <sep>

PaulSnow:
> **Stacked on #1243 (#1232).** This branch starts from `task/1232-seconds-minutes-time-arith`, because both PRs change `EL.g4` and regenerate the ANTLR parser. Two independent regenerations would conflict in every generated file. Until #1243 merges, this PR's diff against main includes its commits (d7fdd2fc, d9ff9087). Only fe8a976c and 10028f4a are #1234. The branch was rebased onto #1243's review fixes. Merge #1243 first, or retarget this PR after it lands.

## What

A new string expression:

```
join names by ", "                 -> "a, b, c"
join (tokenize line by ",") by ","  -> line, unchanged
```

- The result is the elements' string values with the separator between them. Nothing goes before the first or after the last, and an empty array gives `""`.
- It is the inverse of `tokenize`: `join (tokenize s by sep) by sep` is `s`, including empty fields and leading or trailing separators.
- A non-string element contributes its string value, as it would to `+`.
- The separator binds as a single operand. `join names by "|" + "]"` is `(join names by "|") + "]"`, so a join sits inside a concatenation without parentheses: `"[" + join names by "|" + "]"` gives `[a|b]`. A separator built with `+` must be parenthesized: `join names by ("|" + "-")`. `TestJoinExecution` pins all three shapes. This was corrected after review, in 10028f4a; the first version of this text had it backwards.

Implementation:
- Grammar: a `JOIN` token and `JOIN arrayExpr BY strexpr # strJoin` in `strexpr`. The parser was regenerated with the ANTLR 4.13.1 jar.
- Operator: `join` ( array sep -- string ).
- Keyword clash: `join` becomes a keyword. No sample project or fixture uses it as an identifier, and `dtrules verify` passes on every sample project.

## Why

Joining an array needed a loop that left a trailing separator (#1234).

## Tests

- `pkg/dtrules/join_exec_test.go` compiles EL and executes it. It covers:
  - an empty array, one element, three elements
  - an empty separator, integer elements, join inside a concatenation, join of a tokenize
  - the round trip `join(tokenize(s))` for `a,b,c`, `a,,b`, `,a,`, `""` and a string with no separator
- A grammar-sweep row for `strJoin`. The operator is added to `TestAllOperatorsRegistered`.

With the grammar, emitter and operator reverted to the parent commit, `TestJoinExecution` fails ("no viable alternative at input 'join'") and the sweep row fails.

`make check` passes.

## Docs

`dtrules docs el`, `dtrules docs operators` and `docs/el-reference.md`. SPEC.md is unchanged: it does not catalogue EL operators, and the EL reference is where they are specified.

Fixes #1234


PaulSnow (commented):
**[reviewer/fable]** Adversarial review of `task/1234-join-array` @ a7ae9cb1 (only a7ae9cb1 is #1234; d7fdd2fc is #1243, reviewed separately), in throwaway worktrees under /tmp (nothing pushed or changed).

## Verdict

No code defect. `make check` passes on the branch and on `main`+#1243+#1245+#1248 (clean fast-forward on top of #1243; #1248 auto-merges the overrides file with no conflict, and its guard then sees 623 labels and fails when I delete the `strJoin` row). The regenerated parser is byte-identical to regenerating from this branch's `EL.g4` (all 10 files). `TestJoinExecution` and the `strJoin` sweep row fail on `main` (`no viable alternative at input 'join'`). One documentation statement is the inverse of what the grammar does.

## Findings

### 1. The separator does not take "the whole string expression after `by`" — it binds tightly (Low; doc contradicted by grammar)

`docs/el-reference.md:459` and the PR body: "The separator takes the whole string expression after `by`, so parenthesize a join used inside a concatenation: `"[" + (join names by "|") + "]"`."

Measured on the branch (`words` = `{x, y}`, `s` = `"|]"` in the last row):
```
"[" + join words by "|" + "]"   -> "[x|y]"      pf: "[" words "|" join strconcat "]" strconcat
join words by "|" + "]"         -> "x|y]"       pf: words "|" join "]" strconcat
join words by ("|" + "]")       -> "x|]y"       pf: words "|" "]" strconcat join
join words by s + "!"           -> "xx|]yy!"    pf: words s join "!" strconcat
```
So `join a by sep + x` is `(join a by sep) + x`; the parentheses the doc prescribes are unnecessary, and an author who writes `join names by sep + ", "` expecting a two-part separator gets `", "` appended to the result instead. The doc sentence should say the opposite: the separator is a primary; parenthesize the separator if it is an expression. `"[" + join words by ","` with nothing after it gives `"[x,y"`, which is fine.

## Checked and clean

- Keyword clash: scanned every EDD/DT XML, xlsx `sharedStrings`, fixtures, website, legacy and `pkg/` testdata for `join` as an identifier (case-insensitive; grammar is `caseInsensitive`, so `Join`/`JOIN` are reserved too): no hits. `joined` still lexes as IDENT; `set join = "x"` now fails to parse. `dtrules verify` passes on all 11 sample projects with the new binary (the built XML embeds compiled postfix, so every table's EL was recompiled).
- Element types: dates join as their UTC-normalized string (`2026-01-01|2026-01-01T16:30:00Z` — a Chicago 10:30 stamp comes out in Z; consistent with `RDate.StringValue`, worth knowing); entities join as their name (`clock`); `null` contributes `""` (`a||7|true`); a never-set (nil) array is a conversion error, not a panic; integers and booleans as their string values.
- Separators: empty separator over a Unicode string round-trips through `tokenize` (`héllo, wörld`); multi-character separator with empty fields round-trips (`a, b,, c`); the PR's own `a,,b` / `,a,` / `""` cases hold because `strings.Split("")` is `[""]` and `Join([""])` is `""`.
- Round trip: `join (tokenize (join words by ",") by ",") by ";"` composes as expected.
- gofmt clean.

## Not verified

UI vitest (no node_modules); arrays with duplicate-suppression (`dups=false`) — `GetIterator` returns the stored slice so `join` sees whatever the array holds.

PR #1246, opened 2026-09-21
A lone "/" in postfix is integer division, not an empty literal name (#1241)

### LINKED ISSUE #1241: Integer division compiles to a bare '/' that the postfix reader takes as an empty name literal: the result is silently wrong
Found by the #1226 builder and reproduced on DTRules v1.25.0-20-g6a4bc145.

`set result.table = "div:" + string value of (6 / 2) + " mul:" + string value of (6 * 2)` compiles to

```
"div:" 6 2 / cvs strconcat " mul:" strconcat 6 2 * cvs strconcat cvs /result.table xdef
```

and sets `result.table` to **`2 mul:12`**, not `div:3 mul:12`. No error is raised. `set result.num = 1 / 0` also runs with no error and sets nothing.

Cause, per the builder: in the postfix compiler (`pkg/dtrules/compiler/compiler.go`, the `strings.HasPrefix(token, "/")` branch), a lone `/` is read as a literal name (`/` followed by an empty name), not as the division operator. Multiplication is unaffected.

Why the sample projects pass: their divisions have a field or double operand (e.g. `((140 - patient.age) * patient.lean_body_weight) / (patient.pcr * 72)`, `(100.0 * totalGroupIncome) / FPL`), and the emitter then chooses `fdiv`. Only the integer path emits the bare `/`.

Fix: either have the emitter emit a named integer-division operator, or have the postfix reader treat a lone `/` as the operator (a name literal needs at least one character after the slash). Also decide what integer division by zero does; today it is silent.

Regression test: `6 / 2` gives 3; `7 / 2` gives whatever the language defines for integer division; `1 / 0` gives a defined error; each checked through compiled EL.

PaulSnow:
## What

All three postfix readers treated every token starting with `/` as a literal name. A lone `/` therefore became an empty name. The three readers are:

- `pkg/dtrules/compiler/compiler.go`, the object path that the loader and sessions use
- `pkg/dtrules/compiler/bytecode_compiler.go`
- `compileTokenToBytecode` in `pkg/dtrules/repository/repository.go` (XMLRepository → `ruleset.CompiledRuleSet`), which no production path reaches today but which would reproduce this bug if it were wired up Now a token is a literal name only when at least one character follows the slash. A lone `/` falls through to the operator lookup and reaches `opDiv`, which the runtime registers as `/` (with the aliases `div` and `ldiv`).

## Why

Integer division compiles to a bare `/`. That matches `docs/el-reference.md`, which names `/` as the postfix operator for integer division. The reader then misread it, so:

- `"div:" + string value of (6 / 2) + " mul:" + string value of (6 * 2)` gave `2 mul:12`
- integer results came back as non-numbers
- `1 / 0` raised no error

The root cause is in the reader, not the emitter. `/` is a registered operator and can never be a valid literal name, because the name would be empty. Fixing the reader also covers any stored or legacy postfix that uses `/`, and it leaves the emitted postfix unchanged, so nothing has to be regenerated. `/` is the only registered operator whose name starts with a slash, so no other lone-symbol operator had this problem.

The language already defined the behaviour: integer division truncates. `opDiv` uses Go's `/` on int64, which truncates toward zero, and it already raised `Math Exception: Division by zero`. That error was simply never reached. The docs now say "toward zero", describe the division-by-zero error and the lone-`/` rule, and replace an example that showed `fdiv` under "Integer division" with the postfix the compiler actually emits (`count 2 / 1 >=`).

## Stored postfix

I searched every stored `*_postfix` element (XML and JSON) under `sampleprojects/`, `test/` and `pkg/` for a bare `/` token. The only hit is inside a string literal in a TaxReturn policy statement. Nothing needed regenerating.

## Tests

- `pkg/dtrules/authoring/integer_division_test.go` drives the production chain. It authors a table through the SDK (`AddAction` with EL), `Save`s it, loads the XML with the strict loader (`RuleSet.LoadFromDirectory`) and runs `RSession.Execute`. Expected results: the issue's reproducer gives `div:3 mul:12`; `6 / 2` = 3; `7 / 2` = 3; `-7 / 2` = -3; `7 / 2` assigned to a double gives 3.0; `a / 2.0` gives 3.5 (`fdiv`); `1 / 0` and `a / b` with b = 0 fail with `Division by zero`.
- `pkg/dtrules/repository/lone_slash_test.go` checks the same thing for the repository reader.
- `pkg/dtrules/compiler/lone_slash_test.go` checks that both compiler-package readers turn `/` into the division operator (`OpDiv` in bytecode) while `/q` is still a literal name.
- `TestExecute_DivisionByZero_ErrorInTrace` in `hardening_test.go` used to log and pass when no error was raised. It now fails in that case.

Each of these tests fails on origin/main and passes with the fix. I checked this by reverting the reader files, running the tests and restoring the files. `make check` passes.

## Possible overlap with other open branches

- `docs/el-reference.md` is also changed by #1227, #1228, #1229, #1232 and #1234. Their hunks are at lines 438 and later; this PR changes lines 188–190, so the two should not conflict textually.
- No open branch touches `compiler/compiler.go`, `compiler/bytecode_compiler.go`, `repository/repository.go` or `authoring/hardening_test.go`. This PR does not touch `postfix_emitter.go`.

## Not done here

- The bytecode VM's `OpDiv` disagreed with the `/` operator (float division for mixed operands, a different error message). That is #1261: the Go VM is fixed in #1268, and the assembly loop is #1267.
- A table with no conditions and one action column executed no actions in the session. I ran into this while writing the test, which now has a guard condition. I did not investigate it.

Fixes #1241


PaulSnow:
**[reviewer/fable 22:23Z]** Starting adversarial review of this PR. Worktree /tmp/rev-1246 from the PR head (70fbd47b) on top of origin/main 6a4bc145. Plan: read the diff against the issue and docs, run `make check` on the branch and merged onto origin/main, reproduce fail-before/pass-after by reverting the fix under the new tests, then attack the paths the PR body names. Findings will be posted as a review comment here.

PaulSnow:
**[reviewer/fable 22:29Z]** Milestone: `make check` passes on the branch (one commit on top of current origin/main 6a4bc145). Reproduced fail-before/pass-after myself: with compiler.go and bytecode_compiler.go reverted to origin/main, the two lone-slash tests, both integer-division tests and the tightened hardening test fail (`text = "2 mul:12"`, `Execute succeeded (q = )`); restored, all pass. Observation so far: there is a THIRD postfix reader with the old rule, `pkg/dtrules/repository/repository.go:531` (`strings.HasPrefix(token, "/")`); checking whether anything reaches it. Now: stored-postfix sweep, other open PR conflicts, VM OpDiv reachability.

PaulSnow (commented):
## Review (reviewer/fable) — verdict: **merge** (one optional one-liner, see finding 1)

Adversarial review from a throwaway worktree of `task/1241-integer-division` @ 70fbd47b (one commit on top of current `origin/main` 6a4bc145, so the branch result is the merged result). The reader fix is right and complete for the path rule sets actually run through. Two low-severity findings, neither on a shipped path.

### What I ran

- `make check` on the branch: passes. `make check` on a tree with **both** #1246 and #1239 merged onto `origin/main`: passes (35 packages ok).
- `go build -tags archive ./...` / `go vet -tags archive ./pkg/...`: pass (pre-existing `el_parser.go` vet notes only).
- **Fail-before / pass-after, reproduced myself**: with `compiler/compiler.go` and `compiler/bytecode_compiler.go` reverted to `origin/main`: `TestCompile_LoneSlashIsDivision` fails (`"/" compiled to name ()`), `TestCompileToBytecode_LoneSlashIsDivision` fails (`put 1 names in the chunk`, code `[2 6 2 2 202 0]` vs `[2 6 2 2 23]`), `TestIntegerDivision_CompiledELRunsInSession` fails (`text = "2 mul:12"`, `q`/`fq`/`neg` "No Long value exists"), both `TestIntegerDivision_ByZeroIsARuntimeError` cases fail (`Execute succeeded (q = )`), and the tightened `TestExecute_DivisionByZero_ErrorInTrace` fails (`integer divide-by-zero produced no error in the trace`). Restored: all pass. The authoring test goes SDK → `Save` → `LoadFromDirectory` → `RSession.Execute`; nothing is hand-fed.
- **Merge conflicts** against every other open PR (#1235–#1240, #1242, #1243, #1245, #1248): `git merge-tree` clean for all ten. `docs/el-reference.md` is shared with #1235, #1238, #1240, #1243, #1245 — different hunks, as the PR body predicted.

### Can a legitimate name literal be a lone `/`?

No, on the evidence:
- The name layer would accept it (`TryGetRName` at `name.go:63` treats `""` as a valid executable name and strips a leading slash itself), which is exactly why the old reader produced a silent empty name rather than an error.
- The EL emitter never produces it: every `e.emit("/" + …)` in `postfix_emitter.go` (lines 150, 557, 592, 2470, 2478, 2674, 3695, 3701, 3715, 3717) appends a grammar identifier, and `DIVIDE : '/' | 'div' | 'divide'` (`EL.g4:1127`) is the only source of a bare `/`, via `VisitIntMulDiv` → `arithOp` → `mulDivOps`, which picks `/` only when `promote()` gives an integer target (double → `fdiv`, bigint → `b/`, fixed → `fp/`). So `opDiv`'s `LongValue()` on both operands is never fed a double by compiled EL.
- `Register("/", opDiv)` (`operators/math.go:36`) is the only registered operator starting with a slash — the PR's claim holds.
- **Stored postfix sweep** (my own, all `*_postfix` XML elements and JSON under `sampleprojects/`, `test/`, `pkg/`, `legacy/`, `website/`, `docs/`): the only lone `/` is inside the TaxReturn policy-statement string literal the PR names (`TaxReturn_dt.xml:10649`). The StateTax "integer arithmetic" scenarios divide by `10000.0`, so they were already `fdiv`. Nothing needed regenerating.

### Findings

1. **Low — a third postfix reader still has the old rule.** `pkg/dtrules/repository/repository.go:531` (`compileTokenToBytecode`, the `XMLRepository` → `ruleset.CompiledRuleSet` loader) still does `strings.HasPrefix(token, "/")`, so a lone `/` is still an empty literal name there. Reachability: nothing in `pkg/` or `cmd/` imports `pkg/dtrules/repository` (not even a test), so no shipped path reads through it today. But the PR body says "both postfix readers", and if that loader is ever wired up it reproduces #1241 exactly (`6 2 /` → `PushName("")`). Same one-line change (`len(token) > 1 && token[0] == '/'`) would close it; at minimum the PR/docs should not claim the rule is universal.
2. **Low, latent — the bytecode VM's `/` is a different operator.** With the reader fix, the bytecode compiler now emits `OpDiv` for `/` (`bytecode_compiler.go:150`), and `interpreter/vm.go:151-177` implements that as integer division for two integers but **float division** for anything else, with the error `Division By Zero: cannot divide by zero`, whereas the object path's `opDiv` (`math.go:165-186`) is integer-only and raises `Math Exception: Division by zero` — the message `docs/el-reference.md:188` now promises. Reachability: `CompileToBytecode` is called in production only from `cmd/dtrules/main.go:603` with the constant `"1 2 +"`; sessions, `dtrules run`, the interview and API server all compile the object path. So the documented behaviour is what ships. It is still two engines with two answers for one token; worth an issue, not a blocker for this PR.

### Observations (non-blocking)

- `integer_division_test.go:133` calls `state.SetOperatorTable(operators.GetOperatorTable())`; `session.NewSession` and `dtrules run` do not (it serves the bytecode path). Harmless, but it makes the "nothing done by hand" claim slightly less than literal — the object path does not need it.
- The "table with no conditions and one action column executed no actions" note in the PR body is #1230's territory; PR #1237 adds the advisory warning for it.
- The doc change is accurate to the code: `opDiv` uses Go `int64` `/` (truncates toward zero) and the `Math Exception` message matches `math.go:183`.

### Not verified

- The Java/legacy engine under `legacy/` (not built here; only its stored XML was swept).
- The UI's postfix display (`ui/src`) — no slash-prefix tokenizer found by grep, not exercised.
- Whether any *external* project (outside this repo) stores postfix with a lone `/` that relied on the silent empty-name reading.

PaulSnow:
**[reviewer/fable 22:35Z]** Review posted as a comment review: https://github.com/DTRules/DTRules/pull/1246#pullrequestreview-5272490156 — verdict **merge**, with one optional one-liner. Findings: (1) low — a third reader, `pkg/dtrules/repository/repository.go:531`, keeps the old lone-slash rule; unreachable today (nothing imports the package) but the PR's "both readers" claim is incomplete; (2) low/latent — the bytecode VM's OpDiv now backs `/` in the bytecode reader and has different mixed-type and error semantics than opDiv; only reachable from a constant expression in `cmd/dtrules/main.go:603`. Stored-postfix sweep confirms the PR's finding (only the TaxReturn policy-statement string); no legitimate lone-`/` name literal exists. Combined tree with #1239 passes `make check`; clean merge-tree against all other open PRs.

Issue #1247, opened 2026-09-21
Grammar sweep rows don't have to reach the label they cover; 250 of 559 don't


PaulSnow:
Found while fixing #1244 (branch task/1244-grammar-sweep).

The grammar sweep checks that every corpus row **compiles** and emits non-empty postfix. It does not check that the row's DSL **parses through the label the row is filed under**. On main at 6a4bc145, many rows don't.

**Evidence.** I used a throwaway test that parses each merged corpus row (grammar_corpus.tsv plus grammar_overrides.tsv) with the generated parser and walks the tree for the row's label context type. On main, **250 of the 559 merged rows never reach their own label**. 187 of those are not in grammar_known_fails.tsv or grammar_helpers.tsv. Examples:
- `bexpr/boolEntityEq` with `account == account`
- `fexpr/floatFromIndex` with `( double ) indxExpr`: `indxExpr` is a literal identifier here, and `x[1]` parses as `intBytesIndex`. `(fixed) copy of accounts[1]` does reach the `...FromIndex` alternatives.

Two sub-classes:
1. **Stale rows.** 52 corpus keys name labels that are no longer in EL.g4. Examples: `bexpr/boolFloatEq`, `boolIntGt` and the rest of the type-pair comparison family, removed by the numexpr rework (#1148). The coverage guard only checks grammar ⊆ corpus, never corpus ⊆ grammar.
2. **Grammar-unreachable labels.** `localvariables/localLongDefined`, `localDoubleDefined` and `localFixedDefined` can never be chosen. `typedLong`/`typedDouble` and `undefinedIdent` are all `IDENT`, so the parser always takes the `...Undef` alternative. Their rows compile, but through the Undef label. The same may be true of other `typedX : IDENT` pairs.

**Proposed.** The sweep asserts that each row's parse tree contains its label's context type, with known exceptions listed. The coverage guard also fails on corpus keys that are not grammar labels. The unreachable labels are either deleted from the grammar or given a way to be reached.

The 101 rows added in #1244 were each checked this way and reach their labels, except `localFixedDefined`, which is unreachable (its row's note says so).

PaulSnow:
[...]

Issue #1148, opened 2026-08-19
nexpr: one numeric expression rule, so precedence belongs to operators rather than type-pairs


PaulSnow:
Design agreed with Paul (2026-08-19, on #1146). This issue carries the full analysis and the acceptance suite; #1146 stays as the history of how the defect was found and half-fixed.

## The defect, stated precisely

Grouping of same-precedence arithmetic chains in EL is **unstable**: it depends on which grammar alternative ANTLR's adaptive prediction picks for each operator, which in turn depends on surrounding tokens and literal kinds — things that have nothing to do with arithmetic. All measurements below are from main today (post-#1147), verified by compilation:

| expression (fields typed as named) | emits | meaning | C says |
|---|---|---|---|
| `a - b + c` (doubles) | `a b f- c f+` | (a−b)+c | ✓ |
| `a + b - i` (mixed) | `a b f+ i f-` | (a+b)−i | ✓ |
| `a + 2 - b` (int literal) | `a 2 f+ b f-` | (a+2)−b | ✓ |
| **`a + 2.0 - b`** (float literal) | `a 2.0 b f- f+` | **a+(2.0−b)** | ✗ |
| **`the maximum of (a + b - c) and (0.0)`** | `a b c f- f+ 0.0 fmax` | **a+(b−c)** | ✗ |
| `the minimum of (a / b * c) and (d)` | `a b fdiv c fmul d fmin` | (a/b)·c | ✓ (!) |
| **`the maximum of (a - b + c - d) and (0.0)`** | `a b c d f- f+ f-` | **a−(b+(c−d))** | ✗ |
| `the maximum of (i - j + k) and (0)` (ints) | `i j - k + 0 max` | (i−j)+k | ✓ |
| **`-a + b - c`** | `a b f+ c f- fnegate` | **−(a+b−c)** | ✗ |
| `maximum of (a + b - c) and (0.0) > d` | right-grouped | | ✗ |
| `a + b - c > d and i - j + k > n` | left-grouped | | ✓ |

The pattern that finally explains all of it: **iexpr-only chains are always correct; fexpr chains are correct only when prediction happens to route every operator through the same alternative.** A float literal, a construct boundary, or nothing visible at all can reroute one operator and flip the grouping.

## Three mechanisms, all the same root

[...]

PR #1248, opened 2026-09-21
Grammar sweep coverage guard really enumerates EL.g4's labels (#1244)

### LINKED ISSUE #1244: TestGrammarSweep_CoverageGuard extracts 0 labels from EL.g4 and passes vacuously
Found while adding grammar labels for #1232, on main at 6a4bc145.

`TestGrammarSweep_CoverageGuard` (pkg/dtrules/compiler/el/grammar_sweep_test.go) is meant to fail when a labelled alternative in EL.g4 has no corpus row. In practice it cannot fail: `extractGrammarLabels` returns **0** labels.

Evidence: a throwaway test in the package that ranges over `extractGrammarLabels(t)` and logs the count printed `0`. I added 14 new labels in #1232 with no corpus rows, and the guard still passed.

Likely cause: `ruleStartRE` (`^([a-z][A-Za-z0-9_]*)\s*(?:\[[^\]]*\])?\s*:`) expects `name :` on one line. EL.g4 writes the rule name alone on its line and puts the `:` on the next, as in

```
dexpr
    : LPAREN dexpr RPAREN      # dateParen
```

So `rule` is never set, and every `# label` is dropped.

To fix it, track a bare `name` line followed by a `:` line, then run the guard to see which existing labels really lack rows. Every label this change set adds does have rows, in grammar_overrides.tsv.

PaulSnow:
Fixes #1244

## The test was dead (on main, 6a4bc145)
- A throwaway test in the package logged `extractGrammarLabels count = 0`.
- I deleted the `dexpr/dateParen` row from grammar_corpus.tsv. `TestGrammarSweep_CoverageGuard` still passed:
  ```
  === RUN   TestGrammarSweep_CoverageGuard
  --- PASS: TestGrammarSweep_CoverageGuard (0.00s)
  ```
- Cause: `ruleStartRE` wanted `name :` on one line. EL.g4 puts the rule name alone on its line and the `:` on the next, so `rule` was never set and every `# label` was dropped.

## The fix
- The EL.g4 extractor now recognises a bare rule-name line followed by a `:` line, as well as `name :` on one line. It ignores quoted literals such as `'//'` and `'/*'`. A `# label` found outside any rule it recognised is fatal, not skipped.
- A second, independent extraction reads the generated parser. `go/ast` parses el_parser.go, and every `XContext` struct that embeds another `YContext` is taken as label `x` of rule `y`.
- `checkLabelExtraction` requires the two sets to be identical and duplicate-free. Each must hold at least `minGrammarLabels = 500` labels; EL.g4 has 608 on main. A broken extractor, or a stale generated parser, now fails the guard instead of emptying it.
- The compile sweep's compiler gets a stub collection resolver, so that `for all <type> entities` rows compile.

## Labels on main with no row: 101 of 608
With the guard alive, it reported 101 labels without a row: 26 iexpr, 23 dexpr, 14 fexpr, 13 bexpr, 7 numexpr, 6 forallctl, 4 strexpr, 3 localvariables, 2 forallblock, and 1 each for createstatement, eexpr and performstatement. Examples: the whole `dateFirstOf/EndOf{Week,Quarter,Year}[Starting][InZone]` family, `intHourOf`/`intMinuteOf`/`intSecondOf`/`intDayOfWeek`/`intWeekOfYear` and their `InZone` forms, `forallReverse*`, `forallAs*`, `forallTypeEntities*`, `performDynamicTable`, the `boolSameCalendar*` family, `boolNum*`, `fixedFrom*`, and `strFormatDate*`.

This PR adds a row for every one of them in grammar_overrides.tsv. The rows go directly under the header, so they don't collide with #1243/#1245 appending at the end of the file. Each row passes the compile sweep. I also checked with a throwaway parse-tree walk that each row actually parses through its own label, not a sibling. That check is not in this PR.

**One label needs grammar work, not a row:** `localvariables/localFixedDefined` cannot be reached. `typedLong` and `undefinedIdent` are both `IDENT`, so the parser always takes `localFixedUndef`. Its row compiles, but through `localFixedUndef`, and its note column says so. The existing rows for `localLongDefined` and `localDoubleDefined` have the same problem. This is filed as #1247.

## Evidence the guard now works
| Change | Result |
|---|---|
| all rows present | `--- PASS: TestGrammarSweep_CoverageGuard` |
| delete corpus row `dexpr/dateParen` | `FAIL ... grammar coverage gap: 1 labels lack corpus entries` (lists `dexpr/dateParen`) |
| delete new override row `iexpr/intHourOf` | `FAIL ... 1 labels lack corpus entries` (lists `iexpr/intHourOf`) |
| g4 extractor reverted to one-line-only rule starts (the pre-fix behaviour) | `FAIL EL.g4:33: label #emptyAction is outside any parser rule the extractor recognised` |
| parser extractor returns nothing | `FAIL extracted 0 labels from el_parser.go, want at least 500 ...` |
| g4 extractor drops one label | `FAIL EL.g4 and el_parser.go disagree on 1 labels ...` |

## The #1243/#1245 stack
I merged this branch into a throwaway worktree at `origin/task/1234-join-array`, which contains #1232 and #1234. The merge is clean, and both extractors see 623 labels there. `TestGrammarSweep_CompilesAndEmitsPostfix` and `TestGrammarSweep_CoverageGuard` both pass. The stack's new labels all have rows, so the stack will not go red when it lands on top of this.

## Checks
- `make check`: passed (the UI vitest step was skipped because node_modules is not installed).
- `gofmt -l`: clean. `go vet` on the el package reports only the pre-existing unreachable-code findings in the generated el_parser.go, which `make vet` already excludes.

## Found, not fixed (#1247)
- The sweep never checks that a row reaches the label it is filed under. On main, 250 of the 559 merged rows don't; 187 of them are outside known_fails and helpers.
- 52 corpus keys name labels that are no longer in EL.g4, left behind by the #1148 numexpr rework. The guard checks grammar ⊆ corpus only.


PaulSnow (commented):
**[reviewer/fable 22:50Z]** Adversarial review of `task/1244-grammar-sweep` @ 5da583c2 (on current origin/main 6a4bc145), in a throwaway worktree.

## Verdict

**Ready.** I reproduced the dead guard on main and the live guard on the branch, independently re-derived which rows reach their label (100 of the 101 new rows do; the one exception is the one the PR names), tried to think of a grammar change that kills both extractors the same way and could not construct one, and found no conflicts with the other open PRs. Two low-severity notes, no findings that block.

## Reproduced

| where | change | result |
|---|---|---|
| main 6a4bc145 | delete `dexpr/dateParen` from grammar_corpus.tsv | `--- PASS: TestGrammarSweep_CoverageGuard` (dead) |
| branch | same deletion | `FAIL … grammar coverage gap: 1 labels lack corpus entries`, lists `dexpr/dateParen` |

## Do the 101 new rows reach their labels?

I wrote my own walk (not the builder's): parse each merged row with the generated parser the way `compileRow` does (`condition `/`context `/`action ` prefix, the bare-identifier `perform` rewrite, the `;` retry), then look for a `<Label>Context` node anywhere in the tree. Over the 660 merged rows minus helpers and known_fails, **188 do not reach their label** (the builder reported 187 on main; the +1 is the new `localFixedDefined` row). Intersected with the 101 keys added by this PR:

```
localvariables/localFixedDefined      <- the only one; grammar-unreachable, as the PR and #1247 say
```

The other 100 parse through their own alternative. No duplicate keys among the new rows and none collide with a key already on main, so they are real additions rather than overrides.

## Can the guard go silently dead again?

- The two extractors share no code path: one is text + regex over EL.g4, the other is `go/ast` over el_parser.go's struct embedding. A layout change to EL.g4 that the text extractor misses either drops labels (then the sets disagree → fatal) or drops everything (then < 500 → fatal). A stale or differently-generated el_parser.go fails the same way. I could not construct a single edit that moves both sets identically while dropping coverage.
- Residual silent paths are explicit lists, not accidents: a label listed in `grammar_known_fails.tsv` or `grammar_helpers.tsv` counts as covered without compiling; a row can be keyed to a label it never reaches (#1247, filed).
- The floor of 500 against 608 (623 with the #1243/#1245 stack) is a tripwire for "extractor returned almost nothing", not a coverage target; it fails loudly, not silently, if the grammar ever legitimately shrinks past it, and the comment says to lower it then. Sensible.

## Notes (low)

1. `bareRuleRE` accepts `name` or `name[args]` on its own line but not `name returns [...]` / `name locals [...]`. EL.g4 has none today; a future rule written that way would hit the fatal "label outside any parser rule" rather than being skipped — loud, so acceptable, but the message will send someone to the extractor rather than the grammar.
2. `extractParserLabels` lower-cases the first letter to recover the grammar spelling. A label that itself starts with an upper-case letter (`# FooBar`) would come back as `fooBar`, disagree with EL.g4, and fail loudly. Fine as a guard; worth a one-line comment so the next person does not "fix" the extractor.

## Verified

- `make check` passes on the branch (35 packages ok; UI vitest skipped, node_modules absent).
- `TestGrammarSweep_CompilesAndEmitsPostfix` and `_CoverageGuard` pass on the branch; both pass on the branch merged with #1245 (join-array), #1243 (seconds/minutes) and #1246 (integer-division) individually — no textual conflicts with any of #1235–#1248 (`git merge-tree` against every open PR branch).
- The stub collection resolver only affects the compile sweep, not the guard.

## Not verified

- The PR's other numbers (250 of 559 on main, 52 stale keys) — #1247's territory; my count on the merged branch is 188 unreached outside helpers/known_fails, consistent with theirs.
- Whether every new row's DSL is semantically what its label is for (e.g. the `InZone` variants use a real zone). I checked parse reach, not meaning.

PR #1272, opened 2026-09-21
EL: a string joined to a number, date, entity or array concatenates (#1251)

### LINKED ISSUE #1251: String concatenation with a non-string right operand emits nothing ("x" + 1, "x" + current date, ...)
Found under #1247.

## What happens
If a string is joined to anything the grammar doesn't read as a string, the concatenation is dropped. The row still compiles, and only the right-hand side of the comparison is left:

| EL (condition) | postfix |
|---|---|
| `"x" + 1 == "x1"` | `"x1" streq` |
| `"x" + 1.5 == "x"` | `"x" streq` |
| `"x" + current date == "x"` | `"x" streq` |
| `"x" + first of accounts == "x"` | `"x" streq` |
| `"x" + copy of accounts == "x"` | `"x" streq` |
| `"x" + s == "x"` (s a string, for comparison) | `"x" s strconcat "x" streq` |

These go through `strConcatInt`, `strConcatFloat`, `strConcatDate`, `strConcatEntity` and `strConcatArray`. PostfixEmitter has no `Visit` method for any of them. The Go ANTLR base visitor's `VisitChildren` is a no-op, so they emit nothing. visitor_pin_test.go allowlists all five as "dead grammar; base strConcat wins parser-side". That holds only when the right-hand operand is an identifier: an identifier parses as `strXmlValue`, so the base `strConcat` takes it. With a literal, a date or an entity/array expression on the right, the specific alternatives are chosen.

The compile sweep does not catch this, because the postfix as a whole is not empty.

## Repro (main, 6a4bc145)
```go
c := NewCompiler()
out, err := c.CompileCondition(`"x" + 1 == "x1"`)
// out == `"x1" streq`, err == nil
```

## Expected
The left operand is converted and concatenated, as `strConcatName` does (`"x" $nm cvs concat`). Runtime behaviour of the current postfix was not checked; it compares whatever is below the right-hand side on the stack.

PaulSnow:
Fixes #1251

`"x" + 1`, `"x" + 1.5`, `"x" + current date`, `"x" + first of accounts` and `"x" + copy of accounts` parse as `strConcatInt` / `Float` / `Date` / `Entity` / `Array`. None of these had an emitter, so they compiled to nothing. Each now emits `<s> <x> strconcat`, the same operator the base form already uses for a numeric field.

**Before, executed through authoring SDK -> Save -> `RuleSet.LoadFromDirectory` -> `RSession.Execute`:**
- `set calc.text = "x" + 1` compiled to `cvs /calc.text xdef` and failed with `[Stack Underflow] DataPop: Stack is empty  PostFix: ERROR==> cvs`
- the condition `"x" + 1 == "x1"` compiled to `"x1" streq` and failed with a Stack Underflow at `s==`

**After:** `text = "x1"`, `"x1.5"`, `"n=5"`, and both conditions hold.

- The new `pkg/dtrules/authoring/el_runtime_harness_test.go` is a small harness that runs EL through the production chain.
- `strconcat_nonstring_test.go` (authoring) covers run time, and `compiler/el/strconcat_nonstring_test.go` pins the postfix.
- `visitor_pin_test.go` no longer lists the five as dead grammar.
- `docs/el-reference.md` now says the right operand of `+` may be a number, date, entity or array.

**Sample projects:** I recompiled every row of every sample project before and after the fix. Exactly one row changes, `TaxReturn/Calculate_Educator_Expenses` action 1 (`add "  Educator Expense for " + taxpayer.name + ": $" + (the minimum of ...) to job.audit_trail`):
- Before, it compiled to `job.audit_trail swap addto` with the string missing.
- It now builds the string.
- The second commit rebuilds TaxReturn's stored XML with `dtrules build --from-excel`. Only that one row changed. #1269's drift test passes.

`make check` passes.


PaulSnow:

PR #1282, opened 2026-09-21
EL: add to x n / subtract from x n are the values x + n / x - n (#1253)

### LINKED ISSUE #1253: Expression forms "add to X n" / "subtract from X n" emit nothing (intAddTo, intSubFrom, floatAddTo, floatSubFrom)
Found under #1247.

## What happens
The expression forms `add to <ident> <number>` and `subtract from <ident> <number>` emit nothing, as integer (`intAddTo`, `intSubFrom`) and as double (`floatAddTo`, `floatSubFrom`):

| EL | label | postfix |
|---|---|---|
| condition `add to d 5 > 0` | floatAddTo | `0 >` |
| condition `subtract from d 5 > 0` | floatSubFrom | `0 >` |
| context `local fixed f = add to n 5` | intAddTo | `cvfp allocate execute deallocate pop` |
| condition `substring of "abc" from add to n 5 to 2 == "a"` | intAddTo | `"abc" 2 over - substring "a" streq` |

PostfixEmitter has no `Visit` method for any of the four. visitor_pin_test.go allowlists all four as "dead grammar; addtostatement handles `add X to Y`". That holds for the statement `add 5 to n`. It does not hold in an expression position, where these alternatives are the only match. The corpus rows `add to d 5 > 0.0` and `subtract from d 5 > 0.0` already parse through them.

## Repro (main, 6a4bc145)
```go
out, _ := NewCompiler().CompileCondition(`add to d 5 > 0`)
// out == `0 >`
```

## Expected
Either emit the arithmetic (`d 5 +`, typed as intMulBy/intDivBy are via emitMulDivBy), or reject the form at compile time. Which one is intended is not recorded in docs/el-reference.md, which does not list these forms.

PaulSnow:
Fixes #1253

The expression forms `add to <x> <n>` and `subtract from <x> <n>` (`intAddTo`, `intSubFrom`, `floatAddTo`, `floatSubFrom`) had no emitter, so they compiled to nothing. They now emit the value, typed by x's declared type through `emitMulDivBy`, exactly as `multiply x by n` / `divide x by n` already do:
- `add to n 5` -> `n 5 +`
- `add to d 5` -> `d 5 f+`
- `add to f 5` -> `f 5 cvfp fp+`
- `subtract from b 5` -> `b 5 cvbi b-`

x itself is not changed.

**Why a value and not a compile-time rejection** (the issue left this open):
- The Java grammar (`parser.cup`, recovered from git at 12e31a37^) emitted `"/"+n+" "+v+n+" fadd def"`. That defines a name spelled as the number and leaves nothing on the stack, so there was no working Java behaviour to keep.
- The Go port had already made the multiply/divide siblings into values.
- The reasoning is on the issue.

**Before, executed through authoring SDK -> Save -> `RuleSet.LoadFromDirectory` -> `RSession.Execute`, with n = 3 and d = -3.0:**
- `set calc.r = add to n 5` compiled to `cvi /calc.r xdef` and failed with a Stack Underflow.
- `add to d 5 > 0` compiled to `0 >` and failed with a Stack Underflow.
- `substring of "abcdef" from add to n 1 to 6` failed with a Conversion Error.

**After:** r = 8 / -2, rd = 2 / -3.5, text = "ef", and all three conditions hold. n is unchanged.

**Tests:**
- `pkg/dtrules/authoring/add_to_expression_test.go` checks the run-time results.
- `compiler/el/add_to_expression_test.go` pins the postfix for int, double, fixed and bigint.
- `visitor_pin_test.go` no longer lists the four as dead grammar.
- `docs/el-reference.md` gets a section for the four expression forms. The docs test checks its example against the compiler.

**Sample projects:** recompiling every sample row changes 4 action rows in SyntaxTests: Syntax_Examples_3 actions 39/40 and Syntax_Examples_4 actions 23/24 (`set totalIncome = add to totalIncome 75` and similar). The same 4 DSL lines also appear as initial actions in two Run_Test tables.
- Their stored postfix was the broken `cvd /totalIncome xdef`. The second commit rebuilds SyntaxTests' XML with `dtrules build --from-excel`: only those 8 lines changed, and #1269's drift test passes.
- I ran the SyntaxTests and sample-project tests with the old and the rebuilt XML. The results were identical with either XML (40 subtests, 0 failures in both runs). None of the tables those tests execute reaches these rows.

`make check` passes.


PaulSnow:

PR #1286, opened 2026-09-21
EL: a colon or possessive reference makes its entity current (#1257)

### LINKED ISSUE #1257: Colon/possessive references (:e: field, e's field) emit no entitypush/entitypop
Found under #1247.

## What happens
Colon and possessive references (`:account: flag`, `account's flag`) push the entity onto the data stack. They never make it the current entity, so the field is looked up in the wrong scope and the entity is left on the stack:

| EL | postfix |
|---|---|
| condition `:account: flag` | `account flag` |
| condition `account's flag` | `account flag` |
| condition `:account: account == first of accounts` | `account account accounts first cve req` |
| action `set :account: n = 1` | `1 cvi account /n xdef` |
| action `add 1.5 to account's d` (for comparison) | `1.5 account entitypush d f+ /d xdef entitypop` |

The Java grammar these rules were ported from (parser.cup) emits `e entitypush v entitypop` for every colonRef form, rvalue and lvalue, e.g. `colonRef:re RBOOLEAN:b {: RESULT = re+ "entitypush "+b+" entitypop "; :}`. In Go, only `addDestColon` and `subDestColon` do.

The corpus rows for these labels (`boolColonRef`, `entityColonRef`, `dateColonRef`, `leftIexprColon`, ...) are mostly in grammar_known_fails.tsv under #703, because they use the literal word `colonRef`. So the sweep has never compiled them.

## Repro (main, 6a4bc145)
```go
out, _ := NewCompiler().CompileCondition(`:account: flag`)
// out == `account flag`
```
Runtime behaviour was not run.

## Expected
`account entitypush flag entitypop` (the Java emission), for each `...ColonRef` / `left...Colon` visitor.

PaulSnow:
Fixes #1257

`:account: flag`, `account's flag`, `set :account: n = 1` and the other colon/possessive references pushed the entity onto the data stack and never made it current. The field was therefore read from, or stored into, whatever entity was current, and the entity was left on the stack.

Every `XxxColonRef` / `leftXxxColon` visitor now emits `<e> entitypush <field> entitypop pop`.
- The first part is the Java compiler's emission for every colonRef, rvalue and lvalue. I took it from `parser.cup`, recovered from git at 12e31a37^, e.g. `colonRef:re RBOOLEAN:b {: RESULT = re+"entitypush "+b+" entitypop "; :}`.
- The trailing `pop` is needed because this runtime's `entitypop` is `( -- entity )`: it leaves the entity on the data stack. Java's `Entitypop` discarded it.
- Possessive chains (`a's b's n`) used to push `a` and never pop it. They are now balanced, as parser.cup's `possessiveRef` is.

**Before, executed through authoring SDK -> Save -> `RuleSet.LoadFromDirectory` -> `RSession.Execute`.** calc is current; calc.flag=false, calc.n=100; calc.account.flag=true, account.n=7.

| EL | before | after |
|---|---|---|
| condition `:account: flag` / `account's flag` / `:account: flag is true` | false (read calc.flag) | true |
| condition `:account: n == 7`, `account's n == 7`, `:account: n + 1 == 8` | false (read calc.n) | true |
| condition `calc's account's n == 7`, `:calc: account's n == 7` | false | true |
| `set :account: n = 1`, `set account's n = 2` | error: `RIntegerValue: No Integer value exists for entity` | account.n set, calc.n untouched |

**Sample projects.** No sample project's behaviour changes. How I checked:
- I recompiled every row of every sample project. The only rows whose postfix changes are 8 SyntaxTests initial actions: `set InterestingClients = job's clients`, `set client = client's address`, `set effectiveDate = job's effectiveDate`, in Syntax_Examples_2 and Run_Test_5/6/7.
- The second commit rebuilds SyntaxTests' XML with `dtrules build --from-excel`: only those 8 lines changed, and #1269's drift test passes.
- I executed all 23 SyntaxTests tables in fresh sessions, once with the old XML and once with the rebuilt XML. For each table I dumped the error, the data-stack and entity-stack depths, and every field of every reachable entity. The two dumps are byte-identical (13,033 lines each).
- The limit of that evidence: the tables holding those rows fail earlier or later on `validatedCitizenship` being undefined in both runs, so it shows no *observable* change rather than exercising the new form on sample data.

**Tests:**
- `pkg/dtrules/authoring/colon_ref_scope_test.go` checks the run-time results.
- `compiler/el/colon_ref_scope_test.go` pins the postfix, including the chains and the entity-typed field.
- `docs/el-reference.md` gets a "Colon and possessive references" section. The docs test checks its examples against the compiler.

**Not changed here, but found while doing this:**
- `using <e> (<bexpr>)` (VisitBoolUsing, and the other `…Using` visitors that end the same way) emits `entitypop swap pop`, which keeps the entity and drops the result. Executed, `using account (flag)` fails with `BooleanValue: No Boolean value exists for entity`.
- `addDestColon`/`subDestColon` end in `entitypop` with no `pop`, which leaves the entity on the data stack after the statement.

`make check` passes. This PR merges cleanly with #1272, #1279 and #1282, and with task/1229 and task/1249. With all four merged together, `go build ./...` and the authoring, compiler/el and docs tests pass.


PaulSnow:

